PR #4, opened 2021-01-07
More colors, eslint, and Fractal sidebar info

shawnbot:
This PR adds:

1. All of colors in our existing system to the Tailwind theme object
2. eslint with the `standard` config for linting JavaScript 
3. More useful info (with links!) in the Fractal sidebar


src/tokens/colors.js
```diff
@@ -0,0 +1,45 @@
...
+  white: '#FFF',
+  black: '#212123',
+  blue: {
+    1: '#edf4f7',
+    2: '#a9d6ea',
+    bright: '#4f66ee',
+    dark: '#0c1464'
+  },
+  slate: {
+    DEFAULT: '#1c3e57',
```
shawnbot, line 11: @aekong FYI, this is a new-ish feature of Tailwind: the `DEFAULT` key basically defines the `slate` color so that you can use `fg-slate` rather than having to name the default and use, say, `fg-slate-default` or `fg-slate-light`.

src/tokens/colors.js
```diff
@@ -0,0 +1,45 @@
+module.exports = {
+  white: '#FFF',
+  black: '#212123',
+  blue: {
+    1: '#edf4f7',
+    2: '#a9d6ea',
+    bright: '#4f66ee',
```
aekong, line 7: have we talked to design team about renaming these to keep consistent with the numeric labeling (like below `4,3,2,1`)? feels like the right thing to do, but maybe there's a good reason for these adjectives `bright` and `dark`

shawnbot: We haven't, but we should! One thing that's nice about how buttons are written using Tailwind's `@apply` feature is that it will break the CSS build if we reference a utility class that doesn't exist. What I think this means is that we can leave this as-is for now, and if we decide to rename it later we can rely on the build breaking if we've messed up existing components that reference the key we deleted.

PR #5, opened 2021-01-14
Flesh out buttons

shawnbot:
This builds on #4 and sets up component pages for our three button types: [primary](https://design-system-git-buttons.sfds.vercel.app/components/detail/button--default.html), [secondary](https://design-system-git-buttons.sfds.vercel.app/components/detail/button--secondary.html), and [inverse](https://design-system-git-buttons.sfds.vercel.app/components/detail/button--inverse.html).

⚠️ There are, as you can see, some issues with the CSS. Most obviously, the fonts are all wrong (because we're not setting them up properly in the base CSS). @aekong, one idea I'm running with right now is to have three different CSS builds—`base.css` for base styles _only_, `utilities.css` for utilities _only_, and `sfds.css` for both of them—which should give us the ability to start using the utility classes on sf.gov without having to commit to the base styles, which can be trickier to swap in.

A bunch of other things needed to happen in the build setup to support this work:
- The Tailwind configuration needed to be fleshed out to include all of the necessary utilities and theme/token values.
- We're using Babel to compile `src/js/*.js` into `dist/js/`. The [env preset](https://babeljs.io/docs/en/babel-preset-env) coupled with an expansive browserslist config means that we should be able to write modern JS that runs in all the browsers we care about.
- I added a CI workflow to GitHub Actions that lints the JS (but not the CSS! 😱 ) and builds the CSS, JS, and docs (Fractal site) in separate steps so that it's easier to see what failed.
- I upgraded Fractal (and friends) because there's a new version out that fixes some glaring bugs (including one that was making it hard to build the whole project consistently locally, in Vercel, and in GitHub Actions).


aekong (approved):
thanks for the walkthrough 👍

PR #7, opened 2021-01-27
Release 0.0.2: Buttons!

shawnbot:
@aekong and I are continuing to work through publish + composer versioning challenges, which call for a new version to publish. After we figure this out we will work on publishing automatically from PRs (release candidates) and merging to main ("stable" releases).


src/theme.js
```diff
@@ -8,6 +8,11 @@ module.exports = {
   fontSize: {
     body: ['17px', '24px']
   },
+  fontWeight: {
+    normal: 300,
+    semibold: 500,
+    bold: 700
```
shawnbot, line 14: @laurenajong I think we agreed that 500 is "bold", right?

```suggestion
    bold: 500
```

tailwind.config.js
```diff
@@ -86,17 +85,6 @@ module.exports = {
...
-      addVariant('hocus', ({ modifySelectors, separator }) => {
-        modifySelectors(({ className }) => {
-          const prefix = '.' + e(`hocus${separator}${className}`)
-          return suffixes
-            .map(suffix => `${prefix}${suffix}`)
-            .join(',\n')
-        })
-      })
-    })
+    require('tailwindcss-interaction-variants')
```
shawnbot, line 88: @aekong The custom plugin that I added for the "hocus" (hover + focus) variant wasn't working correctly, so I brought in [this package](https://github.com/benface/tailwindcss-interaction-variants). The only thing to be aware of is that its "hocus" variant doesn't include classnames to activate the hover/focus styles (it only gives us the pseudo-classes), so I had to explicitly style `.hover` and `.focus` with the state-specific utilities in the CSS [here](https://github.com/SFDigitalServices/design-system/blob/ebf361a215b8ebe9b70d5ee4a5e52af4c4d68a36/src/css/components/button.css#L16-L19), [here](https://github.com/SFDigitalServices/design-system/blob/ebf361a215b8ebe9b70d5ee4a5e52af4c4d68a36/src/css/components/button.css#L28-L31), and [here](https://github.com/SFDigitalServices/design-system/blob/ebf361a215b8ebe9b70d5ee4a5e52af4c4d68a36/src/css/components/button.css#L41-L44).

src/css/components/button.css
```diff
@@ -0,0 +1,55 @@
...
+    @apply appearance-none;
+    @apply cursor-pointer;
+    @apply font-semibold;
+    @apply rounded;
+    @apply px-2 py-1;
+    @apply bg-blue-bright;
+    @apply text-white;
+    @apply border-0 border-solid border-current;
+    @apply text-center justify-center items-center;
+    @apply whitespace-nowrap;
```
shawnbot, line 13: @laurenajong this is what I did to prevent buttons from wrapping. We can't prevent authors from making the content wrap explicitly, but I think this is good enough for now.

src/theme.js
shawnbot: We [discussed this in Slack](https://sfdigitalservices.slack.com/archives/CEERV5DCG/p1616106655003800) and agreed to punt on renaming our font weights for now.

aekong:
Not sure if this was addressed or not, but will we ever need a tertiary button? For example, "yes", "no", "cancel"?

laurenajong:
Buttons themselves look good! I don't fully understand what is the behavior that makes it not wrap -- I'm sure what you have is good, but maybe you could explain it briefly during mid-sprint review.

Also, will a designer ever need to know or care about the class names? Can also talk about that in mid-sprint review.

PR #8, opened 2021-03-18
Create typography and spacing styles

shawnbot:
For DESSYS-22. Includes spacing scales and some other bits from #10.

This PR takes a first pass at implementing the typography styles from [this Figma document](https://www.figma.com/file/KFrLq31jDbhxYnWtsnEH5C/Typography-and-Spacing-(New!)?node-id=427%3A11). You can review them [with the Fractal UI](https://design-system-git-typography-sfds.vercel.app/components/detail/typography.html) or [without](https://design-system-git-typography-sfds.vercel.app/components/preview/typography.html).

I've minted class names that match with what we call our type styles in Figma:

Type style | Class | Notes
:--- | :--- | :--
Display large | `display-lg`
Display small | `display-sm`
X-large title | `title-xl`
Large title | `title-lg`
Medium title | `title-md`
Small title | `title-sm`
X-small title | `title-xs`
Big description | `big-desc`
Body | `text-body` | This is a Tailwind built-in, but we can alias it to `body` for consistency if need be
Small text | `small`

I've brought in [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono) for monospace text, which we may not need in the system but will very likely want in Fractal.

Per @laurenajong's suggestion, there's also a dynamic display of the font-size, line-height, and letter-spacing below each specimen's title (in the left column):

Small screen | Wide screen
:--- | :---
image | image


src/theme.js
```diff
@@ -34,5 +20,17 @@ module.exports = {
...
+    '1/2': '50%',
+    '2/2': '100%',
+    '1/3': '33.3333333%',
+    '2/3': '66.6666666%',
+    '3/3': '100%',
+    '1/4': '25%',
+    '2/4': '50%',
+    '3/4': '75%',
+    '4/4': '100%',
+    full: '100%'
```
shawnbot, line 34: I needed to stub out the `width` scale after gutting the `spacing` scale because the Tailwind default is to inherit spacing into the width scale.

coreyhunt:
Looks good. Matches the Figma specs.

PR #9, opened 2021-03-26
Component markup

aekong:
* updates the nunjucks path
* separates button markup into its own file and include in button macro

@shawnbot the purpose of this was to have a single button markup file that we could use for import into other templates on the sf.gov drupal theme side.

i haven't figured out how to get just the markup to show up in the fractal static site, though, so it can be a standalone reference for expected markup for the button component.

tried to add in another panel in the mandlebrot theme, but think it requires customizing the theme itself to pick up a tab like "markup"

### DIFF
--- a/fractal.config.js
+++ b/fractal.config.js
@@ -20,7 +20,7 @@ fractal.set('project.title', pkg.project.title)
 fractal.set('project.package', pkg)
 
 const nunjucks = require('@frctl/nunjucks')({
-  paths: ['src/templates'],
+  paths: ['src/components'],
   filters: {
     iconify (content) {
       if (content && content.icon) {
--- a/src/components/button/button.html
+++ b/src/components/button/button.html
@@ -1,4 +1,4 @@
-{% macro button(properties) %}
+{% macro button(label, prefix, suffix, class) %}
   {% include "./button/markup.html" %}
 {% endmacro %}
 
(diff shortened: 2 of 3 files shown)


fractal.config.js
```diff
@@ -20,7 +20,7 @@ fractal.set('project.title', pkg.project.title)
 fractal.set('project.package', pkg)
 
 const nunjucks = require('@frctl/nunjucks')({
-  paths: ['src/templates'],
+  paths: ['src/components'],
```
shawnbot, line 23: Sorry about this. I added it so that we can override the default Fractal theme's templates, in which case we'll want both of them:

```suggestion
  paths: [
    'src/components',
    'src/templates'
  ],
```

shawnbot (commented):
I appreciate the intent, but I'm not sure if we're ready for this. We haven't thought through the button component API, so these templates are not really ready for prime time. I think we're better off trying to incorporate the button styles into sf.gov first and develop Twig templates that make sense there before we settle on an API.

When I say "API", I'm really just thinking in terms of what the Twig/Nunjucks template variables might look like. Nesting `prefix.icon` and `suffix.icon` probably don't make sense here or in sf.gov; maybe we want `icon` and `iconPosition` (`before` or `after`?), or maybe (probably) we want different templates/components for each layout? 🤷 

Icons are something that we will probably need to do differently on sf.gov, too. The Fractal site currently uses an old and experimental build of the `<sfgov-icon>` custom element with SVG icons baked in, and which is definitely not ready for prime time yet.

Anyway, what I guess I'm saying is that we should probably build the templates we need in sf.gov first, refine them there, then extract those templates into something that we can use elsewhere (like formio?) once we're sure they work well.

aekong:
yeah that makes sense, thanks for the input.  i'll close this pr for now and we can revisit when we're closer to fleshing out the templates.  thanks!

PR #10, opened 2021-04-02
Stub out spacing

shawnbot:
This is what my proposal from yesterday looks like if we decide to "name" our spacing scale with the actual pixel values. After having to tweak the padding classes I think this is actually way clearer and more predicable: the previous `px-2` utility adds 20px of horizontal padding, but you wouldn't know that unless without internalizing the spacing scale, looking it up, or guessing that I'd made that choice the first time around. `px-20` just does exactly what it says it will, and I think that's a good thing.

There's a [component page](https://design-system-git-spacing-sfds.vercel.app/components/detail/spacing.html) for the spacing utilities, but I'm honestly not sure how useful it is.


aekong (approved):
i like when i don't have to math 👍

src/tokens/spacing.js
```diff
@@ -0,0 +1,23 @@
...
+    4,
+    8,
+    16,
+    20,
+    24,
+    32,
+    40,
+    60,
+    80,
+    100
```
shawnbot, line 13: @aekong @Styler415 These are what I have for the spacer sizes right now. Am I missing anything?

PR #13, opened 2021-04-08
Button fixes

shawnbot:
This PR fixes some issues with my first pass on button styles:

1. `.btn` now gets `display: inline-flex` instead of `display: flex`. This allows them to "shrink-wrap" to their contents' width. Adding `.btn-block` makes it `display: flex` and full-width.
2. Buttons now get `text-decoration: none` via the `no-underline` Tailwind utility.
3. Buttons _and_ the `*` selector (defined in our base styles) now both get `box-sizing: border-box` so that padding works correctly.

The [button component template](https://design-system-git-button-fixes-sfds.vercel.app/components/detail/button--default.html) also now renders both inline and block variations of both a link (`<a>`) and a `<button>` element with the `btn` class so that we can spot-check them. The rendering is not great, so let's fix that before merging:

image

@aekong You can test this by installing `sfgov-design-system@0.0.0-12f9409`. If this works for you, let's start a release branch for it and merge to that.

### DIFF
--- a/src/components/button/button.html
+++ b/src/components/button/button.html
@@ -11,6 +11,21 @@
 {% endmacro %}
 
 <div class="p-4 h-screen {{ wrapper.class }}">
+  <div class="mb-2">
+    <h2>Links</h2>
+    <a href="javascript:void(0)" class="{{ class }}">This is a link</a>
+    <br>
+    <a href="javascript:void(0)" class="{{ class }} btn-block">This is a block link</a>
+  </div>
+
+  <div class="mb-2">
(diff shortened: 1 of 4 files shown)


aekong:
looks good! 

resolves the two full width thingies:

- https://pr-886-sfgov.pantheonsite.io/departments/city-administrator/office-civic-engagement-and-immigrant-affairs
- https://pr-886-sfgov.pantheonsite.io/departments/city-administrator/office-civic-engagement-and-immigrant-affairs

🚢

aekong (approved):

PR #16, opened 2021-05-17
Bundle fonts

shawnbot:
This PR looks at using [Fontsource](https://fontsource.org/docs/getting-started) packages to import the `@font-face` declarations and files for the fonts we use: [Rubik](https://fonts.google.com/specimen/Rubik) and [Noto Sans TC](https://fonts.google.com/specimen/Noto+Sans+TC) (and [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono) for monospace). Not loading fonts from the Google Fonts CDN gives us a lot more flexibility and can drastically improve our load time and render performance.

There's one other big change here. Rather than having two different font stacks (one for Latin and another for Chinese) and specifying where each one is used, we combine the fonts into our default "sans" stack, combine Latin and Chinese characters wherever we want, and let the browser figure out which font has a glyph for each one.

This approach only works out of the box if we're okay with how these characters interact. Here's are some comparisons using snippets from the [Moscone Center South location page](https://sf.gov/zh-hant/location/moscone-center-south-vaccine) (left) and some hacked [specimens](https://design-system-267whhgym-sfds.vercel.app/components/detail/typography.html) (right):

Before | After 
:--- | :---
image | image
image | image

The Rubik file sizes are decent:

```
 36K    dist/css/files/rubik-latin-300-normal.woff2
 36K    dist/css/files/rubik-latin-400-normal.woff2
 36K    dist/css/files/rubik-latin-500-normal.woff2
108K    total
```

The Noto Sans TC ones are... not great. We could probably do some magic to separate Chinese characters for critical CSS (i.e. "中文" for the language link) and defer loading the rest:

```
1.4M    dist/css/files/noto-sans-tc-chinese-traditional-300-normal.woff2
1.4M    dist/css/files/noto-sans-tc-chinese-traditional-400-normal.woff2
1.4M    dist/css/files/noto-sans-tc-chinese-traditional-500-normal.woff2
4.2M    total
```

The other thing we need to ensure is that the Latin subset of Rubik includes all of the characters that we need for Spanish, Filipino, and other languages that we may use on sf.gov. IIRC, Latin subsets typically include [Latin-1 supplement](https://unicode-table.com/en/blocks/latin-1-supplement/) characters such as ñ, ü, ø, ç, ¡ and ¿; but if we ever intend on hosting, say Russian text then we'll need to include additional subsets for Cyrillic glyphs.


DecliningLotus:
> The Noto Sans TC ones are... not great. We could probably do some magic to separate Chinese characters for critical CSS (i.e. "中文" for the language link) and defer loading the rest.

I would suggest referring to the weight files directly - https://github.com/fontsource/fontsource/blob/main/packages/noto-sans-tc/400.css

As you can see in there, the font files are split into even smaller subsets to accommodate exactly what you are looking for - it's something done to the font files of very large languages. You just need to find the specific unicode characters you want to prioritise and then find the corresponding subset font file that matches that unicode-range in the CSS file.

DecliningLotus:
@shawnbot, if you mean bundling as in something like using Webpack, you'll be completely fine! The browser will handle everything due to the CSS unicode-range feature. That is the preferred way of loading fonts and mirrors exactly what Google Fonts does.

You should look into benchmarking the difference in downloads from the browser rather than what's getting copied into your dist folder, unless your repo does actually have a different purpose.

DecliningLotus:
> What I'm trying to wrap my head around is whether/how to preload these fonts when we don't know which Chinese characters are going to show up above the fold on a given page.
I would leave the CSS as is and let the browser figure it out. As for preloading... it's something that I haven't really gotten about investigating for a while now (fontsource/fontsource#83). My main concern for that is Webpack renaming the font files with a content hash, which makes it difficult to ascertain what link to use for preloading the font.

In your case though, it's as simple as preloading `css/files/noto-sans-tc-119-400-normal.woff2` for example since you're just copying the files directly into your public directory. It is recommended to preload only your critical font files, since your page will remain in a blocking state till the files you have instructed to preload are fully downloaded. Any Chinese characters that you can't predict will show up, don't preload it. 

 I _do_ know that we'll need to display the "中文" language link, and the characters for those unicode codepoints ([4E2D](https://www.compart.com/en/unicode/U+4E2D) and [6587](https://www.compart.com/en/unicode/U+6587)) both live in the `noto-sans-tc-119-*` font files. So I guess if we figure out which weights are critical we could preload those weights of Rubik (Latin) and _just_ the 119 subset of Noto Sans TC?

Yup. That sounds about right.

DecliningLotus:
Also to help cut down the size of the font bundle even more, I would suggest switching to [Variable Fonts](https://fontsource.org/docs/variable-fonts) for Rubik which should be a third in size compared to what you are working with now.

Unfortunately, Noto Sans TC doesn't support that yet, else you could've saved a huge amount of bandwidth with that.

PR #23, opened 2021-06-10
Release 1.1.0

shawnbot:
### Pull requests
- [x] #24
- [x] #25 (included in above)

### 🚀 Features
- We have a new monospace text style that uses Roboto Mono
- Our typography CSS now includes styles for `<kbd>` elements
- There is a new CSS distribution (`dist/css/fonts.css`) that includes all of the necessary `@import` rules for our Google-hosted custom fonts.

#16 explores bundling the font files directly and removing the Google Fonts dependency. We're not quite ready for that, but I think it's something that we should look as we introduce Noto Sans TC formally.


aekong (approved):
🚢 !

laurenajong:
Can you coordinate with Zakiya and see how including #27 will impact https://sfgovdt.jira.com/browse/SG-1412?

PR #24, opened 2021-06-10
Add mono font

shawnbot:
This PR adds a `mono` entry to the `fontFamily` of our Tailwind theme. Preview links:

- [Typography](https://sfgov-design-system-pr-24.herokuapp.com/components/detail/typography.html) ([standalone](https://sfgov-design-system-pr-24.herokuapp.com/components/preview/typography.html))
- [kbd component](https://sfgov-design-system-pr-24.herokuapp.com/components/detail/kbd.html) ([standalone](https://sfgov-design-system-pr-24.herokuapp.com/components/preview/kbd.html)) from #25 

The "font stack" is just `Roboto Mono` and the `monospace` [generic name](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#generic-name). We can add other fallbacks in between if we have a preference for other system fonts, like [this one](https://systemfontstack.com/) or [this one](https://qwtel.com/posts/software/the-monospaced-system-ui-css-font-stack/#the-monospaced-system-ui-css-font-stack).

This PR also introduces a new `fonts.css` bundle that includes all of the requisite Google Fonts `@import` rules. For now, the font imports are _not_ included in `sfds.css` (which [bundles](https://github.com/SFDigitalServices/design-system/blob/mono-font/src/css/sfds.css) base styles, utilities, and components). But maybe it should? 🤔


coreyhunt:
This should be 14/18 instead of 16/24. It is correct in #25

src/css/fonts.css
```diff
@@ -0,0 +1,3 @@
+@import url('https://fonts.googleapis.com/css?family=Rubik:300,400,500&display=swap');
+@import url('https://fonts.googleapis.com/css?family=Roboto+Mono:400&display=swap');
+@import url('https://fonts.googleapis.com/css?family=Noto+Sans+TC:300,400,500&display=swap&subset=chinese-traditional');
```
shawnbot, line 3: I think that until we introduce Chinese type specimens (as in #16) we might want to leave this out. We're not even referencing it anywhere in our `fontFamily` Tailwind configuration yet, and that will make it easier to benchmark the performance impacts of adding it.

```suggestion
```

aekong (approved):
😍

PR #22, opened 2021-05-27
Move icons into this repo

shawnbot:
[👀 Preview][preview]

I'm moving the code over from the (private, archived) [repo](https://github.com/SFDigitalServices/sfgov-icons) from which the `<sfgov-icon>` custom element is currently published. I don't think it makes sense for it to live outside of this repo for now. This should be part of a new minor release (#51).

- [x] Add `<sfgov-icon>` custom element definition to `sfds.js` bundle
- [x] Add [Rollup] to our JS build setup (and keep [Babel] as our compiler)
- [x] Stub out [icons documentation][preview]
- [x] Output static SVG files in our npm distribution
- [x] Code review from @jacksongonzales 
- [x] Get preliminary design sign-off from @coreyhunt
- [x] Initial QA of [icon library](https://sfgov-design-system-pr-22.herokuapp.com/resources/icon-library/) and [icon usage](https://sfgov-design-system-pr-22.herokuapp.com/usage/icons/) pages

### `sfds.js` bundle
This PR adds a new JS bundle, `dist/js/sfds.js`, which currently only defines the `<sfgov-icon>` custom element and the icon SVG as of June of 2020. (I haven't copied over the scripts to update the SVG data for them, but we can do that in a follow-up PR.)

### Rollup
Previously we were only using [Babel] to _compile_ our JavaScript source, which did not reference any external dependencies. Now we're using [Rollup] with Babel to _bundle_ external dependencies with our own code then _compile_ the resulting bundles into browser-friendly JS. Our custom element doesn't rely on external libraries, but I took this opportunity to replace some JS libraries hosted on unpkg.com with locally built bundles (specifically, the [`<clipboard-copy>` custom element](https://github.com/github/clipboard-copy-element) is now included in our `docs.js` bundle).

#### ESM bundles
Every bundle we build now also generates an ESM-friendly file ending in `.mjs`, which Node.js understands natively. (MJS = Module JavaScript, and ESM = ECMAScript Modules, which are basically interchangeable.)

[Rollup]: https://rollupjs.org/guide/en/
[Babel]: https://babeljs.io/
[preview]: https://sfgov-design-system-pr-22.herokuapp.com/foundations/icons/

PR #27, opened 2021-06-24
Add keyboard section title, make it responsive

shawnbot:
For DESSYS-79.

Preview: https://sfgov-design-system-pr-27.herokuapp.com/components/detail/kbd.html

@coreyhunt and @laurenajong I've sketched out a way that we could make this responsive, but it will probably need another round of design tweaking because the spacing isn't quite right. One thing I've done to prevent the `<kbd>` elements from wrapping awkwardly is reducing their margins at smaller screen widths (e.g. with `mx-4 lg:mx-8` utilities, which translates to: "4px horizontal margins, 8px on large screens").

Small screen | Large screen
:--- | :---
image | image

@aekong This uses the [slate a11y SVG icon](https://sf.gov/themes/custom/sfgovpl/src/img/icons/accessibility-slate.svg) from sf.gov in an `` so we can't adjust the color like we do with `<sfgov-icon>` (because I haven't added that icon to it yet; #22 needs to be merged before we can add it), but I think this works as a placeholder for now. The thing to note is that I've given the image a `height` to set the maximum height it will scale to and `width` to ensure that it pushes the title far enough to the right in their parents' flex box.


src/components/kbd.html
```diff
@@ -1,18 +1,28 @@
 <div class="p-20 text-slate max-w-lg">
+
+  <div class="p-16 lg:p-40 rounded bg-blue-1">
+    <div class="flex items-start lg:items-center mb-20">
```
shawnbot, line 4: `items-start lg:items-center` here aligns the icon and title at the top ("start") by default, but centers them on larger screens because it looks better when they're on one line.

src/components/kbd.html
```diff
@@ -1,18 +1,28 @@
...
+      <img
+        class="flex-shrink-0"
+        src="https://sf.gov/themes/custom/sfgovpl/src/img/icons/accessibility-slate.svg"
```
shawnbot, line 7: This URL will need to change on sf.gov, and should reference an icon with the same color as the text.

src/components/kbd.html
```diff
@@ -1,18 +1,28 @@
...
+        width="24" height="20" alt="">
+      <h3 class="title-sm m-0 ml-8 lg:ml-16">
```
shawnbot, line 9: `ml-8 lg:ml-16` gives the icon and title more breathing room (from 8 to 16px) in larger viewports.

src/components/kbd.html
```diff
@@ -1,18 +1,28 @@
...
+        Navigating dashboards with a keyboard
+      </h3>
+    </div>
+    <ul class="list-none space-y-20 m-0 p-0 lg:ml-40">
```
shawnbot, line 13: * `space-y-20` puts 20px of space between each of its children
* `lg:ml-40` is tuned to align the list items with the title above, and will need to change if the icon `width` changes

src/components/kbd.html
```diff
@@ -1,18 +1,28 @@
...
+      <li><kbd class="mr-4 lg:mr-8">Control</kbd> + <kbd class="mx-4 lg:mx-8">Enter</kbd> to enter the dashboard</li>
+      <li><kbd class="mr-4 lg:mr-8">Tab</kbd> or <kbd class="mx-4 lg:mx-8">Arrow</kbd> to move between visuals</li>
+      <li><kbd class="mr-4 lg:mr-8">Control</kbd> + <kbd class="mx-4 lg:mx-8">Right arrow</kbd> to enter a visual or filter</li>
+      <li><kbd class="mr-4 lg:mr-8">Escape</kbd> to exit a visual or dashboard</li>
```
shawnbot, line 17: The `m*-4 lg:m*-8` pattern here creates responsive margins around the `kbd` elements: 4px on small screens, 8px on large. You could also wrap the `+` text in `<span class="inline-block mx-4 lg:mx-8">` and adjust the `kbd` element margins accordingly, or simplify them all down to `mx-4`.

zakiya (commented):
I'm not sure what the plan is for integrating this component into Drupal but if this was a twig file the main concern would be that the the text is hardcoded so therefore not editable in Drupal or human translatable.

PR #31, opened 2021-07-15
Add pre-publish checks

shawnbot:
This should help guard against accidentally deleting CSS bundles in the future. [This build](https://github.com/SFDigitalServices/design-system/pull/31/checks?check_run_id=3079608198#step:11:18) is an example of what it will look like if we (accidentally or intentionally) remove any of the CSS bundles that we have in there currently.

If we add a new bundle in the future, we should add it to the [`dist/css/{...}.css` glob](https://github.com/SFDigitalServices/design-system/blob/735cbcbde46e4165174d705ec577bf8dd693f1e8/scripts/prepublish-checks#L20). If we remove a bundle in the future, we will _have_ to remove it from the same glob to get the checks passing again.


aekong:
hella cool 😍

aekong (approved):

PR #45, opened 2021-09-01
refactor typography tokens to be mobile first

jacksongonzales:
I ended up leaving the `letterSpacing` values untouched, just to avoid bulking up the `typography.js` file, and since we only ever use two values: `-1px` and `-2px`. If folks think it'd make the tokens self-document better, I'm happy to make the declarations for each token more explicit.

Also, as another alternative, it looks like there's a shorthand for hitting all three text attributes (`fontSize`, `lineHeight`, `letterSpacing`) by doing e.g.:
```
fontSize: {
    'display-lg': ['44px', {
        letterSpacing: '-1px',
        lineHeight: '48px'
    }]
}
```

### DIFF
--- a/docs/foundations/typography.md
+++ b/docs/foundations/typography.md
@@ -54,28 +54,28 @@ with color banner) or display [large](#large-title) (pages with no color
 banner) to style the one-time page title (`<h1>`).
 
 #### X-large title
-> <div class="title-xl">{{ specimens.default }}</div>
+> <div class="title-xl-desktop">{{ specimens.default }}</div>
 
 #### Large title
-> <div class="title-lg">{{ specimens.default }}</div>
+> <div class="title-lg-desktop">{{ specimens.default }}</div>
 
 #### Medium title
-> <div class="title-md">{{ specimens.default }}</div>
+> <div class="title-md-desktop">{{ specimens.default }}</div>
 
 #### Small title
-> <div class="title-sm">{{ specimens.default }}</div>
+> <div class="title-sm-desktop">{{ specimens.default }}</div>
 
 #### X-small title
-> <div class="title-xs">{{ specimens.default }}</div>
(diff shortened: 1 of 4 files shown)


jacksongonzales:
@shawnbot agreed with the class names and putting the styles just for the docs in `docs.css`. Should I do the same with the desktop token names? E.g. `@apply text-desktop-title-lg` to `@apply text-title-lg-desktop`

jacksongonzales:
@shawnbot 
> @jacksongonzales
> 
> > agreed with the class names and putting the styles just for the docs in `docs.css`. Should I do the same with the desktop token names? E.g. `@apply text-desktop-title-lg` to `@apply text-title-lg-desktop`
> 
> I think the token names are good as-is. The only thing that I think needs to move are all of the `.*-desktop` and `.*-mobile` CSS declarations. Does that make sense?

Sorry, phrased clunkily; I meant, do you think we should change the token names in the 'variant of X' style you mentioned? Currently the desktop ones are prefixed `desktop` rather than suffixed

aekong:
i'm in agreement with the suffix "variants" for mobile and desktop.  

> these styles probably shouldn't go into the published bundles but should instead live in docs.css so that it's really clear they're not to be used

... _yet_, right? because at some point we will use these typography styles when we're ready to integrate with sf.gov?

jacksongonzales:
@aekong I believe this is just for classes we're declaring explicitly for the Typography docs page, to get around the problem of displaying a mobile style in a non-mobile viewport and vice-versa. Basically they're workarounds for responsive design, unless I'm mistaken 🙃

shawnbot (approved):
This looks great! Would you mind adding the `-desktop` suffix to the rest of the examples in `typography.md` before merging? Then we can add the mobile variants in that branch.

PR #49, opened 2021-09-29
Add responsive container and grid styles

shawnbot:
This is a first pass at implementing the grid system in our CSS (DESSYS-45) with @jacksongonzales.

### Utilities
We're leaning heavily on Tailwind's utilities. The grid container needs the [`grid`](https://tailwindcss.com/docs/display#grid) display utility (to set `display: grid`), and then some combination of the following utilities and their responsive variants:

1. [`grid-cols-*`](https://tailwindcss.com/docs/grid-template-columns) establishes a column "template", in which the `*` can be 1, 2, 3, 6, or 12. In other words, the number of columns available.
2. [`grid-rows-*`](https://tailwindcss.com/docs/grid-template-rows) similarly sets up a row "template" with 1, 2, or 3 rows. (By default, rows are "automatic" and the grid will grow vertically to accommodate however many rows are occupied.)
3. [`gap-*`](https://tailwindcss.com/docs/gap) utilities establish spacing between columns and/or rows, using our existing spacing scale. For instance, `gap-28` spaces cells by 28px in both dimensions.

The grid elements ("cells") slot into the column and row templates by default, but can grow to occupy multiple columns and/or rows with:

1. [`col-span-*`](https://tailwindcss.com/docs/grid-column#spanning-columns) utilities, in which the `*` can be any integer from 1 to 12 (e.g. `col-span-1`, `col-span-6`, `col-span-12`), control the number of columns that the element spans. This should be less than or equal to the number of columns available.
2. [`row-span-*`](https://tailwindcss.com/docs/grid-row#spanning-rows) utilities can be used to make grid cells span 1, 2, or 3 rows. This is crucial to building the responsive layouts that our design system docs site uses.

### New classes
We have two new classes that you can combine, either in a single element or to one or more ancestors, that apply the grid system specced in the [Figma grid]:

1. `responsive-grid` applies the `grid`, `grid-cols-*` and `gap-*` utilities and their responsive variants
2. `responsive-container` applies horizontal margin and `max-width` utilities to position a grid container (or really, any "container") in the viewport with the proper spacing on its left and right edges.

Building a responsive grid, then, is really just a matter of combining these new classes with utilities to customize the grid behavior. For instance, a traditional sidebar layout might look like:

```html
<div class="responsive-container responsive-grid grid-cols-3">
  <div class="col-span-1">side bar</div>
  <div class="col-span-2">content area</div>
</div>
```

And you could make it responsive by adding variants (the `lg:` prefix applies those utilities when the viewport width is ≥ the "lg" breakpoint of 1076px):

```html
<div class="responsive-container responsive-grid grid-cols-1 lg:grid-cols-3">
  <div class="lg:col-span-1">side bar</div>
  <div class="lg:col-span-2">content area</div>
</div>
```

You can preview a more complete example of this skeletal layout [on CodePen](https://codepen.io/shawnbot/pen/jOLbpXB?editors=1100):

Mobile | Widescreen
:--- | :---
image | image

<details>
<summary> 🕥 Background</summary>

### Flexbox
First we tried just enabling [gap utilities](https://tailwindcss.com/docs/gap) in Tailwind so that we could space out flexbox elements evenly (without the fussiness of responsive padding and margins). Unfortunately, proportional column widths that add up to 100% don't take the gap into account, so while a 3-column layout [works for one row](https://codepen.io/shawnbot/pen/RwgdxJo):

image

...it breaks as soon as you add another box and need it to [wrap to the next row](https://codepen.io/shawnbot/pen/YzQgYdG):

image

### CSS Grid
Next we looked at Tailwind's [grid template column](https://tailwindcss.com/docs/grid-template-columns) utilities, as this seemed to match what we were looking for: a grid container element that tells its children how big to be and how to wrap by default, but also allows for children to size and position themselves. [Et voilà](https://codepen.io/shawnbot/pen/ExXMdvK):

image

I have some more to write about the implementation here, but you can see a bunch of examples on the [spacing page](https://sfgov-design-system-pr-49.herokuapp.com/foundations/spacing/) and a few others [on CodePen](https://codepen.io/shawnbot/pen/QWgXbxx?editors=1000).
</details>

[css grid]: https://css-tricks.com/snippets/css/complete-guide-grid/
[figma grid]: https://www.figma.com/file/svEAhwa9aGljDbcKfOCBsq/Layout?node-id=862%3A98


src/tokens/spacing.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     48,
     60,
     80,
+    96,
     100
```
shawnbot, line 16: @coreyhunt It feels weird to me to have both 96 and 100 here, but it's more of a smell than anything else. Thoughts?

src/theme.js
```diff
@@ -35,26 +35,54 @@ module.exports = {
     3: '3px',
     4: '4px'
   },
+  gap: {
+    ...spacing
+  },
+  gridColumn: {
+    'span-auto': 'auto',
+    ...Object.fromEntries(
```
shawnbot, line 43: TIL about [Object.fromEntries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) 🚀

src/theme.js
```diff
@@ -35,26 +35,54 @@ module.exports = {
...
+  },
+  height: {
+    auto: 'auto',
+    full: '100%',
+    ...spacing
+  },
+  margin: {
+    auto: 'auto',
+    ...spacing,
+    'gutter-auto': `calc((100vw - ${breakpoints.lg}) / 2)`
```
shawnbot, line 66: @jacksongonzales I think we might not need this anymore. We should be able to use `lg:px-96` + `mx-auto` to achieve the same results.

shawnbot: Er, I guess `lg:mx-96 xl:mx-auto`? 🤔

jacksongonzales: did we try `mx-auto` on xl viewports and it didn't work? I can't remember 🤔

shawnbot: I went back into this and saw that I broke `mx-auto` in cda62fbc605b474b6456763735305fb2f76045d0 by adding a `margin` key that didn't list `auto` explicitly (the [default](https://tailwindcss.com/docs/margin) is the spacing scale + `auto`). So I think this is fixed, but it's worth testing to be sure.

jacksongonzales:
@shawnbot I'm realizing it might be to our benefit to add some of the grid-row utilities along with columns. When experimenting with the implementation [here](https://sfds-grid-demo.glitch.me/) you'll notice that the left sidebar only extends as far as its content on the y-axis. Since we have no row configuration, it only knows to let each container take up one row.

I was able to find a way to make the sidebar render correctly with just columns, though it feels like it's leaning towards fighting the grid utility when there's a within-pattern tool we could be using. Do you think bringing in [rows](https://tailwindcss.com/docs/grid-row#customizing) could be useful outside of the DS website?

src/theme.js
```diff
@@ -35,26 +44,72 @@ module.exports = {
...
+        n,
+        `repeat(${n}, minmax(0, 1fr))`
+      ])
+    )
+  },
+  gridTemplateRows: {
+    ...Object.fromEntries(
+      gridRows.map(n => [
+        n,
+        `repeat(${n}, minmax(0, 1fr))`
```
jacksongonzales, line 82: I'm not sure if we should make a more custom class for site purposes, but I had to modify this to e.g. `2: repeat(2, 0fr)` so that a multi-row layout wouldn't be filled with a buncha white space. might be something we wanna discuss, lmk!

jacksongonzales: actually, might make more sense to do `n: repeat(n, minmax(0, auto))`?

src/tokens/spacing.js
coreyhunt: Is this related to the spacing that Scott created [here](https://www.figma.com/file/svEAhwa9aGljDbcKfOCBsq/Layout?node-id=760%3A195)?
Because if it is, you would be correct, and I see a few other things that are incorrect if that is true.

shawnbot: Yup. Looks like 24 and 48 should go, too?

shawnbot: Er, not 24. 24 is needed to make boxes that have the same line-height as body text. At least, I _think_ it should be in there... but we haven't used it yet. So maybe not. 😬

src/css/components/grid.css
```diff
@@ -0,0 +1,12 @@
+.responsive-grid {
+  @apply grid grid-cols-6;
+  @apply gap-20;
+  @apply md:grid-cols-12 md:gap-28;
```
shawnbot, line 4: These utilities implement the grid settings in [Figma](https://www.figma.com/file/svEAhwa9aGljDbcKfOCBsq/Layout?node-id=862%3A98):
* `grid-cols-6` establishes a 6-column grid on small screens
* `gap-20` sets the gap ("gutter") between columns to 20px
* `md:grid-cols-12` switches to a 12-column grid on tablet viewports and larger
* `md:gap-28` increases the gap ("gutter") to 28px on tablet viewports and larger

We've chosen to separate these settings from the responsive margins so that users of the system can choose which elements implement the "internal" (grid columns and gutters) and "external" spacing (horizontal margins and max-width).

jacksongonzales:
> @jacksongonzales:
> 
> > this looks good; I'm pretty sure we've got what we need for implementation on the website! the [glitch](https://glitch.com/edit/#!/sfds-grid-demo?path=index.html%3A177%3A43) is only missing correct row/column alignment for the tablet view. It'll be easier (I think) to implement once we're actually working in the eleventy templates — at least as it is in the example, it involves doing a buncha bulky "show this `div`/hide this `div`" based on breakpoints which I'm hoping we can avoid in the real thing
> 
> I think we should give that a shot in this PR. Want to pair on it some time on Friday?

remembering we never got around to this — should we work on the actual site implementation in a separate PR?

src/css/components/grid.css
```diff
@@ -0,0 +1,12 @@
...
+}
+
+.responsive-container {
+  @apply mx-20 max-w-lg;
+  @apply md:mx-28;
+  @apply lg:mx-96;
+  @apply xl:mx-auto;
```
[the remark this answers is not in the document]
shawnbot: I'm going to cherry-pick f85e566 to this branch for the change from `max-w-lg` to `max-w-xl`.

docs/_includes/example.njk
```diff
@@ -1,6 +1,6 @@
 {% set button_class = 'btn btn-inverse rounded-4 px-8 py-4' %}
 <div class="my-20 border-1 border-solid border-slate-4">
-  <div class="p-20">
+  <div class="{{ meta.wrapper_class | default('p-20') }}">
```
shawnbot, line 3: This change makes it possible to "pass" `wrapper_class="..." after the `html` language specifier in a code example block to modify the class of the wrapping `<div>` in the HTML output. This is useful for some previews that may need margins instead of padding or to eliminate the outer whitespace altogether.

jacksongonzales (approved):
Looks good! Let's Get This Bread*

*Released

PR #52, opened 2021-11-03
Implement new layout + grid breakpoints

jacksongonzales:
This PR puts the utilities and new classes from #49 into practice, in addition to implementing the new [layout/design](https://www.figma.com/file/089UadxdkLJ0xP5vAWj5V2/SF-Design-System-Site-Design?node-id=335%3A2894) of the site ([DESSYS-146](https://sfgovdt.jira.com/browse/DESSYS-146)).

### Layout
The core structure of the site lives in `base.njk`, which we've refactored to use grid utilities. Throughout, we've also applied breakpoints to the new classes to make the site responsive. We've taken a 'mobile-first' approach, leading to patterns that look like this:

```html
<div class="col-span-full lg:col-span-9 {% if toc %} xl:col-span-6 {% endif %}">
  {% block content %}
    {% include 'main.njk' %}
  {% endblock %}
  {% include 'source-meta.njk' %}
</div>
``` 
The classes above say "Let the following content span the full length of columns by default; on large screens, only span 9 (of 12) columns; on extra-large screens where the table of contents is visible, span 6."

### New site features
1. We've added a toggle-able dialog menu for small/medium viewports where, previously, documentation content would be competing for visible space with the navigation. This is leveraging github's `[details-dialog-element](https://github.com/github/details-dialog-element)` plugin.
2. To see a more literal illustration of the grid utilities in action, one can now press `Control + G` to toggle a grid overlay that mimic Figma's [layout grid feature](https://help.figma.com/hc/en-us/articles/360040450513-Create-layout-grids-with-grids-columns-and-rows#h_5d2961fd-4edb-49e1-96f0-057197a82014). This is leveraging a [hotkey](https://github.com/github/hotkey) plugin.


package.json
```diff
@@ -79,5 +79,8 @@
     "unist-util-select": "^3.0.4",
     "unist-util-visit": "^2.0.3",
     "watchy": "^0.9.9"
+  },
+  "dependencies": {
+    "@github/details-dialog-element": "^3.1.3"
```
shawnbot, line 84: This should be a dev dependency until we've decided how to incorporate it into the published components.

src/tokens/widths.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * These are "ideal" content widths that accommodate
+ * nice line lengths and evenly sized grids with standard
+ * spacing (20 or 28px) *and*
```
shawnbot, line 4: Not sure how I meant to finish this... 🤔

docs/_data/assets.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
   scripts: [
     { src: '/dist/js/docs.js' },
     { src: 'https://unpkg.com/@sfgov/icons@0.0.1/dist/sfgov-icons.umd.js', defer: true },
-    { src: 'https://unpkg.com/@github/clipboard-copy-element@1.1.2/dist/index.umd.js', defer: true }
+    { src: 'https://unpkg.com/@github/clipboard-copy-element@1.1.2/dist/index.umd.js', defer: true },
+    { src: 'https://unpkg.com/@github/details-dialog-element@3.0.4/dist/index.umd.js' }
```
shawnbot, line 11: Heads up: this will change after #22 merges because we're no longer importing from unpkg. Instead, we'll add this to `src/docs.js`:

```js
import '@github/details-dialog-element'
```

Then install it as a dev dependency:

```sh
npm i -D @github/details-dialog-element@3.0.4
```

src/css/components/grid.css
```diff
@@ -5,7 +5,7 @@
 }
 
 .responsive-container {
-  @apply mx-20 max-w-lg;
+  @apply mx-20 max-w-xl;
```
shawnbot, line 8: I was scratching my head over why the layout felt so cramped, and discovered that we had the max-width of `.responsive-container` clamped to the `lg` width rather than `xl`. We may need to pair on this to get it 100% right because I think this means that there will be no margin when the viewport is 1076px wide, but at least we'll be seeing the container's intended width.

jacksongonzales: nice, good catch; looks much better now

docs/_data/assets.js
jacksongonzales: sweet, I'll address this (along with using the actual menu icon) once #22 is merged in

src/css/components/grid.css
jacksongonzales: looks like it maintains the `mx-96` at 1076px and above 🎉

shawnbot: Nice! One thing I'm wondering about is a grid overlay on the site a la Figma. Maybe with [hotkey](https://github.com/github/hotkey) and a `<details>` element? 🤓 

```html
<details>
  <summary
    data-hotkey="Control+g"
    aria-label="press Control+g to enable the grid overlay"
  ></summary>
  <div class="z-50 responsive-container responsive-grid fixed h-full">
    {% for i in range(1, 12) %}
      <div class="col-span-1 bg-red-1"></div>
    {% endfor %}
</details>
```

docs/_includes/base.njk
```diff
@@ -1,37 +1,80 @@
...
 
-{% block main %}
-  <div class="px-20 max-w-xl mx-auto mt-40 mb-100">
-    <div class="lg:flex justify-end">
-      <div class="lg:w-1/4 lg:mr-60 mb-40">
-        {% block nav %}
-          {% include 'nav.njk' %}
-        {% endblock %}
+<div class="bg-grey-1">
+  {% set overlayClasses = 'z-50 pointer-events-none responsive-container responsive-grid fixed h-full top-0 left-0 right-0 opacity-75' %}
```
shawnbot, line 16: This is great! I think we could ditch this line though, and just wrap the two overlays in a single element.

```suggestion
```

docs/_includes/base.njk
```diff
@@ -1,37 +1,80 @@
...
-          {% else %}
-            {% include 'main.njk' %}
-          {% endif %}
-        {% endblock %}
-        {% include 'source-meta.njk' %}
+      <div class="hidden md:grid {{ overlayClasses }}">
+        {% for i in range(0, 12) %}
+          <div class="bg-red-1"></div>
+        {% endfor %}
       </div>
```
shawnbot, line 34:
```suggestion
```

docs/_includes/base.njk
```diff
@@ -1,37 +1,80 @@
...
+      <summary
+        data-hotkey="Control+g"
+        aria-label="press Control+g to enable the grid overlay"
+        class="text-slate-light font-mono"
+      >Press <kbd>CTRL</kbd> + <kbd>G</kbd> to toggle grid overlay</summary>
+      <div class="md:hidden {{ overlayClasses }}">
+        {% for i in range(0, 6) %}
+          <div class="bg-red-1"></div>
+        {% endfor %}
       </div>
```
shawnbot, line 29: I think this would work:

```suggestion
      <div class="z-50 pointer-events-none responsive-container fixed h-full top-0 left-0 right-0 opacity-25">
        <div class="responsive-grid grid md:hidden">
          {% for i in range(0, 6) %}
            <div class="bg-red-3"></div>
          {% endfor %}
        </div>
        <div class="responsive-grid hidden md:grid">
          {% for i in range(0, 12) %}
            <div class="bg-red-3"></div>
          {% endfor %}
        </div>
      </div>
```

Note the `opacity-25` and `bg-red-3` changes; I think that looks more like Figma:

image

shawnbot (changes requested):
I love the grid overlay! I just made a couple of suggestions to tweak it, but otherwise this is looking great and we should merge it next week.

docs/_includes/base.njk
```diff
@@ -1,37 +1,80 @@
...
-        {% endblock %}
-        {% include 'source-meta.njk' %}
+<div class="bg-grey-1">
+  <div class="responsive-container flex justify-between py-8">
+    <details>
+      <summary
+        data-hotkey="Control+g"
+        aria-label="press Control+g to enable the grid overlay"
+        class="text-slate-light font-mono"
+      >Press <kbd>CTRL</kbd> + <kbd>G</kbd> to toggle grid overlay</summary>
```
shawnbot, line 22: Nit: I think the word "Control" is more meaningful here than `CTRL`:

```suggestion
      >Press <kbd>Control</kbd> + <kbd>G</kbd> to toggle grid overlay</summary>
```

docs/_includes/base.njk
```diff
@@ -1,37 +1,80 @@
...
+      {% include 'repo-links.njk' %}
+    </div>
+  </div>
+</div>
+
+<div class="responsive-container flex lg:hidden mt-60 mb-28 items-center">
+  <details>
+    <summary>
+      <sfgov-icon class="pr-20" symbol="menu"></sfgov-icon>
+    </summary>
```
shawnbot, line 46: This needs an `aria-label` and/or `title` to be legible by AT. I think we want to label the `<summary>` rather than the icon, but in practice it probably doesn't matter. I'm not really sure what the accessible label should be: "Open menu"?

shawnbot (approved):
This looks great to me. Nice work! 💪

src/theme.js
```diff
@@ -93,7 +103,7 @@ module.exports = {
     ...spacing
   },
   maxWidth: theme => ({
-    ...breakpoints,
+    ...widths,
     ...theme('width')
```
shawnbot, line 107: I think these do the same thing? 🤔 

```suggestion
    ...theme('width')
```

src/theme.js
```diff
@@ -2,6 +2,16 @@ const breakpoints = require('./tokens/breakpoints')
...
+const widths = require('./tokens/widths')
+
+// available column grid templates
+const gridColumns = [1, 2, 3, 6, 12]
+// possible per-column spans
+const gridColumnSpans = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+// available row grid templates
+const gridRows = [1, 2, 3]
+// possible per-row spans
+const gridRowSpans = gridRows
```
shawnbot, line 14: Bad merge, sorry! 🤦 

```suggestion
```

PR #54, opened 2021-12-07
Docs fixes and more

shawnbot:
There's a lot going on here, all in service of making the docs code example and clipboard copy experiences better:

1. I've consolidated all of the Nunjucks stuff into a single file, `lib/nunjucks/index.js`. Previously this stuff lived in `lib/eleventy`, which was awkward. I've updated the Eleventy config to also use the whole custom _environment_ (filters, globals, functions, and whatever else we need) rather than just the custom filters.

2. I've reworked the code sample template to (mostly) match @coreyhunt's [Figma design](https://www.figma.com/file/089UadxdkLJ0xP5vAWj5V2/SF-Design-System-Site-Design?node-id=625%3A4310), including tooltips for the CodePen and copy buttons. We also now have both the new "clipboard" and "code" icons!

3. While building out the disclosure controls in the original Figma design, I came up with some nice patterns for styling `<details>` elements with our utilities. This led to the creation of a custom `details` Tailwind variant that allows us to conditionally style open disclosures a la sf.gov. I wrote these up in the new [Details component page](https://sfgov-design-system-pr-54.herokuapp.com/components/details/).

4. I added the ability to set a fixed height on code sample iframes in the code block meta, like so:

    ```
    ```html height="64"
    ```

    This is necessary when the calculated height of the iframe content is shorter than its dynamic height, which hides content when you open and close `<details>` elements.

5. I added a `tooltip()` template macro (which really just adds a bunch of classes to a `<div>`) and put that into use on both the [Icon library page](https://sfgov-design-system-pr-54.herokuapp.com/resources/icon-library/) and both the CodePen and copy buttons in code samples.

6. I rewrote the "Copied" tooltip text behavior without all of the fancy opacity fading. Now, you just click (or press <kbd>Enter</kbd> while focused) the copy button, the tooltip text changes to "Copied", its `aria-hidden` attribute is set to `false` so that screen readers will announce it, and only when the focus moves does it disappear and get reset to its original text.

7. I fixed up the blame URLs in the "source meta" footer of each file, but I think the commit URLs are wrong when they're fetched from GitHub for some reason. 🤔 

8. I replaced jsdom with cheerio to parse headings from content and generate the tables of contents. It's much faster now, so eleventy doesn't output warnings about it taking too long.

9. I throttled the forced reloading of `.eleventy.js` so that eleventy reloads are faster and less flaky.


.eleventy.js
```diff
@@ -23,10 +23,7 @@ module.exports = config => {
     startLevel: 3
   })
 
-  for (const [name, filter] of Object.entries(filters)) {
-    config.addFilter(name, filter)
-  }
-
+  config.setLibrary('njk', environment)
```
shawnbot, line 26: For some reason we weren't already doing this, which means that our globals weren't available in the Nunjucks templates that eleventy renders (as opposed to the ones that we render in remark 🙃 ). Either way, it should only be _adding_ stuff, not removing it.

docs/_includes/example.njk
```diff
@@ -1,66 +1,84 @@
...
     <iframe
       class="w-full border-0 bg-white"
-      onload="{{ resize_handler }}"
-      onresize="{{ resize_handler }}"
+      {% if meta.height %}
+        height="{{ meta.height | e }}"
+      {% else %}
+        onload="{{ resize_handler | e }}"
+        onresize="{{ resize_handler | e }}"
+      {% endif %}
```
shawnbot, line 13: Demoing `<details>` elements necessitated this change. If the height of the element is measured when it's closed, then the content that becomes visible when you click the summary is hidden because the iframe doesn't grow to accommodate it. Currently we don't have any way to resize the iframes in response to interactions, so we'll probably run into this with tooltips and other interaction patterns.

docs/_includes/example.njk
```diff
@@ -1,66 +1,84 @@
...
       {% else %}
         srcdoc="{{ code | e }}"
       {% endif %}
     ></iframe>
   </div>
-  <div class="bg-grey-1 border-0 border-t-1 border-solid border-slate-4 relative">
-    <div class="flex justify-start">
-      <div class="p-20 flex-auto" id="{{ id }}">
+  <div class="bg-grey-1 border-0 border-t-2 border-solid border-slate-4 relative">
+    <details class="details-reset p-20"{% if meta.closed != 'true' %} open{% endif %}>
```
shawnbot, line 24: The [Figma design](https://www.figma.com/file/089UadxdkLJ0xP5vAWj5V2/SF-Design-System-Site-Design?node-id=625%3A4310) shows code examples closed initially. I left a comment there for Corey about whether we want this to be the default, or if it's okay to have them open by default and closable.

lib/remark/inline-code.js
```diff
@@ -6,7 +6,7 @@ const colorPattern = /^#[a-f0-9]{3,6}$/i
 module.exports = function inlineCode () {
   return (tree, file) => {
     visit(tree, 'inlineCode', (node, index, parent) => {
-      addClass(node, 'inline-flex bg-grey-2 rounded-2 px-4 py-2')
+      addClass(node, 'inline-code')
```
shawnbot, line 9: Previously we were applying classes in two places: here and in `docs.css`. With this change, we just target `.inline-code` in `docs.css`.

lib/tailwind/details.js
```diff
@@ -0,0 +1,14 @@
+const plugin = require('tailwindcss/plugin')
+
+module.exports = plugin(({ addVariant, addUtilities, e }) => {
+  addVariant('details', ({ modifySelectors, separator }) => {
+    modifySelectors(({ className }) => {
+      const suffix = e(separator) + className
+      return [
+        `details[open].open${suffix}`,
+        `details[open] > summary.open${suffix}`,
+        `details[open] > summary .open${suffix}`
```
shawnbot, line 10: Our `details` variant utilities can target:
1. An open `<details>` element,
2. The _directly_ descendent `<summary>` of an open `<details>`, and
3. Any descendent of the directly descendent `<summary>` in an open `<details>`

So:

```html
<details class="details-reset open:bg-red-2 p-8">
  <summary class="p-8 open:bg-yellow-2">
    I'm transparent (red) when closed, yellow when open
  </summary>
  <div class="p-8 mt-8 bg-white">
    I'm always white because I'm only visible when the details is open
  </div>
</details>
```

src/css/base.css
```diff
@@ -23,4 +23,15 @@
...
+  .details-reset {
+    summary {
+      list-style: none;
+
+      &::marker,
+      &::-webkit-details-marker {
+        display: none;
+      }
+    }
+  }
```
shawnbot, line 36: Is this a base style, a utility, or something else? 🤔

src/css/typography.css
```diff
@@ -47,6 +47,14 @@
     @apply text-title-md font-medium;
   }
 
+  .title-sm {
+    @apply text-title-sm font-medium;
+  }
+
+  .title-xs {
+    @apply text-title-xs font-medium;
+  }
```
shawnbot, line 56: I cherry-picked 451531fb398541bdd05d516f1d7305c2a4d12ddb from #52 to get `title-sm` and `title-xs` classes working.

tailwind.config.js
```diff
@@ -56,18 +56,14 @@ module.exports = {
     'padding',
     'pointerEvents',
     'position',
+    // TODO [^3]: add preflight (https://tailwindcss.com/docs/preflight)
     'margin',
     'maxWidth',
-    'opacity',
```
jacksongonzales, line 61: we're still using opacity for the grid overlay. though it appears to be working in the deployment still 🤔

jacksongonzales (commented):
could you apply the `details-reset` class to the grid toggle in `base.njk`? it's currently showing the lil carrot

docs/components/details.md
```diff
@@ -0,0 +1,108 @@
...
+  </summary>
+  <div class="mt-8 bg-blue-1 rounded p-20">Details</div>
+</details>
+```
+
+<div class="flex items-start gap-20 bg-blue-1 rounded p-20 my-40">
+  <sfgov-icon symbol="info" class="mt-2"></sfgov-icon>
+  <div>
+    It's not necessary to label icons with <code>aria-label</code> or
+    <code>title</code> attributes. Doing so may confuse screen reader users.
```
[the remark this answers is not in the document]
shawnbot: Thanks for flagging. Would you mind putting a link to her audit here?

docs/components/details.md
```diff
@@ -0,0 +1,108 @@
...
+## Putting it all together
+This example uses border, color, typography, and spacing utilities with `open:`
+variants to style a `<details>` element as they're currently shown on SF.gov:
+
+```html height="180"
+<details class="details-reset rounded overflow-hidden border-solid border-3 border-action">
+  <summary class="
+    title-xs bg-white text-action
+    open:bg-action open:text-white
+    hocus:bg-action hocus:text-white
```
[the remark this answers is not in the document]
shawnbot: Yeah, after messing with it I agree. I nixed the `hocus:` variants to match what's going on on SF.gov (e.g. at the bottom of [this page](https://sf.gov/see-if-you-can-add-new-units-your-residential-property?nid=610)), so we can punt on making a decision about it.

tailwind.config.js
shawnbot: This appears to be a GitHub bug 😱 . It's [not actually missing](https://github.com/SFDigitalServices/design-system/blob/9aa2820cd1e1b8fd138fb8ae15e78f12b3a89c47/tailwind.config.js#L55), at least as of 9aa2820.

docs/components/details.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+## Open variants
+We provide `open:` utility class variants so that you can conditionally
+show and hide elements within the `<summary>` when the `<details>` element is
+open (whenever the `open` attribute is present).
+This is how you would visually swap a
+<sfgov-icon symbol="chevron-right" aria-label="chevron-right icon"></sfgov-icon>
+when closed for a 
+<sfgov-icon symbol="chevron-down" aria-label="chevron-down icon"></sfgov-icon>
```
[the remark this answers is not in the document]
shawnbot: Here's what they look like as of 9aa2820cd1e1b8fd138fb8ae15e78f12b3a89c47, after introducing an `::icon{symbol=...}` directive that upgrades to `<sfgov-icon>` and wrapped them in `<code>` elements: 

image

It's not _great_, but it's passable IMO.

docs/components/details.md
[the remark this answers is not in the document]
shawnbot: I've removed this little box for now. We can revisit the examples once we have clearer guidance on whether icons need labels.

tailwind.config.js
shawnbot: False alarm! Turns out they just weren't sorted properly 🙃

docs/components/details.md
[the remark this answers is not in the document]
jacksongonzales: nice! def an improvement

jacksongonzales (approved):
woo! looks great 🙂

PR #55, opened 2021-12-09
Lint CSS; add expiring-todo-comments rules for JS + CSS

shawnbot:
There are a couple of things happening here, the overarching goal being to lint comments in our JS and CSS that mark bits of code for fixing, deprecation, etc. _in a specific version_. The original art is from [unicorn/expiring-todo-comments], one rule from a [popular eslint preset](https://github.com/sindresorhus/eslint-plugin-unicorn) that looks for comments like this:

```js
const foo = 'bar' // TODO [>=3]: delete this
```

The [plugin docs explain conditions](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md#conditions) well, but the one I'm interested in is package versions, which helps us mark bits of code for modification and **reminds** us (by failing our builds! 😬) that we intended to do something by the version specified. The general pattern of the comment text is:

```
TODO [semver]: reason
```

The rule checks if the `version` in `package.json` "satisfies" the `semver` bit (a specific version or [semver range] pattern), and if so, raises a linting error with the `reason`. So, parts of code that we want to deprecate in a future (major) version should look like:

```js
module.exports = {
  spacing, // TODO [>=2.3]: spacing scale should include line heights
  // TODO [>=3]: delete
  letterSpacing: { ... }
}
```

If we bump the package version (without deleting or updating the comments) to `2.3.0` the first one will fail, and at `3.0.0` the second will. 🎉 

### Stylelint
I've also brought in [stylelint] to lint our CSS, and built a _very_ basic [`local/expiring-todo-comments` rule][local/expiring-todo-comments] for flagging `TODO [semver]: reason` comments in our stylesheets as errors. This was also a good opportunity to fix some minor stylistic issues in our stylesheets. 🎉 

⚠️ **NB:** I've also upgraded Eleventy to 1.0.1 in this PR and extracted the focus ring color as a token (`colors.grey.focus`?), so this is technically a feature branch.

[unicorn/expiring-todo-comments]: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md#readme
[stylelint]: https://stylelint.io/
[semver range]: https://devhints.io/semver
[local/expiring-todo-comments]: https://github.com/SFDigitalServices/design-system/blob/313cc9040b5e23668d1b519cdb501ab960f7b1a1/lib/stylelint/expiring-todo-comments.js


.stylelintrc.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+  extends: [
+    'stylelint-config-standard'
+  ],
+  plugins: [
+    './lib/stylelint/expiring-todo-comments'
+  ],
+  rules: {
+    'at-rule-no-unknown': [true, {
+      ignoreAtRules: [
```
jacksongonzales, line 10: ugh, so excited for this. looking at our css files as they are now makes me ANXIOUS

src/css/base.css
```diff
@@ -16,8 +16,8 @@
 
   :focus,
   .focus {
-    outline: 2px dotted #aeb0b5;
-    outline-offset: 3px; /* TODO [^3]: remove outline-offset (or set to 0) */
+    outline: 2px dotted theme('colors.grey.focus');
+    outline-offset: 3px; /* TODO [>=3]: remove outline-offset (or set to 0) */
```
jacksongonzales, line 20: silly question maybe, but how does the linter know this is referring to the tailwind package?

jacksongonzales: oh nvm, it's the dessys version 😵‍💫

jacksongonzales (approved):
looks awesome, excited to put this to use!

src/css/base.css
shawnbot: Yeah, if we want to support `[dependency@version]` we'll need to figure out how to extract that matching logic more completely from the eslint rule.

PR #56, opened 2021-12-09
Lint markdown

shawnbot:
This PR adds [markdownlint] to our collection of linting tools, a basic [configuration] with some good rules to start us off, and a `lint-docs` npm script that runs it on all of the `.md` files in the `docs` directory. I've also created a [custom rule] that ensures each of our markdown files has YAML front matter (the `---` bits at the top) and at least a `title:` key, which is essentially required by our templates.

- [x] install [markdownlint]
- [x] add a basic [configuration]
- [x] add a [custom rule] to enforce YAML front matter and `title:`
- [x] install and test the GitHub Actions [problem matcher] to add annotations for markdown linting errors (example: [this annotation])
- [x] remove testing violations

This is branched off #55, but could be rebased if we want to merge it separately.

[markdownlint]: https://github.com/DavidAnson/markdownlint#readme
[configuration]: https://github.com/SFDigitalServices/design-system/blob/6ddff56ebc6a20be8121d14e2b3732ac0b6ae344/.markdownlint.yml
[custom rule]: https://github.com/DavidAnson/markdownlint/blob/main/doc/CustomRules.md
[problem matcher]: https://github.com/xt0rted/markdownlint-problem-matcher


lib/markdownlint/frontmatter.js
```diff
@@ -0,0 +1,47 @@
+const yaml = require('js-yaml')
+
+module.exports = {
+  names: ['front-matter'],
+  description: 'YAML frontmatter',
+  tags: ['local'],
+  function: (params, onError) => {
+    const { config, frontMatterLines } = params
+    const { required_keys: requiredKeys = [] } = config
```
jacksongonzales, line 9: is this pulled from `.markdownlint.yml`? and do we need to write an explicit linting function checking for the title?

shawnbot: Yup, it's here: https://github.com/SFDigitalServices/design-system/blob/b8367f3a205b0997dd2a7f8773b5f3140d0eadb3/.markdownlint.yml#L5-L7

shawnbot: And to answer your other question: No, I don't think we're quite ready to enforce any rules about page titles :)

jacksongonzales (approved):
this look good to merge into #55 ! I'll probably run thru and do some testing on that branch for all the new linting features

jacksongonzales:
> Thanks. I could also `git rebase -i` and just get the relevant commits so this can merge into #92 in either order.

that works too, would probably make reading #55 easier

PR #59, opened 2021-12-17
fix(docs): typo in grid styles and add nav for resources

jacksongonzales:
Correcting a typo in the inlined styles for the grid overlay. We stray from using design tokens once and look what happens!

### DIFF
--- a/docs/_includes/base.njk
+++ b/docs/_includes/base.njk
@@ -8,6 +8,7 @@ layout: skeleton
   {{ nav_link('/foundations') }}
   {{ nav_link('/components') }}
   {{ nav_link('/usage') }}
+  {{ nav_link('/resources') }}
 {% endset %}
 
 {% set toc = content | toc %}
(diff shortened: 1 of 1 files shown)


docs/_includes/base.njk
```diff
@@ -8,6 +8,7 @@ layout: skeleton
   {{ nav_link('/foundations') }}
   {{ nav_link('/components') }}
   {{ nav_link('/usage') }}
+  {{ nav_link('/resources') }}
```
shawnbot, line 11: Let's nix this from this PR and get the grid styles working here. I'm queueing up a 2.2.1 release for docs fixes.
```suggestion
```

shawnbot (changes requested):

docs/_includes/base.njk
shawnbot: (I think Resources is supposed to come before Usage anyway)

shawnbot (approved):

PR #62, opened 2021-12-21
Add group-hover variant

aekong:
Adding `group-hover` variant to facilitate [styling based on parent state](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state).

For something like this:

|default|hover (white border)|
|---|---|
|Screen Shot 2021-12-21 at 12 37 46 AM|Screen Shot 2021-12-21 at 12 37 54 AM|


aekong:
nope that was it! sorry i missed that.  thanks!

PR #66, opened 2021-12-22
Typography: set "medium" font-weight to 600

shawnbot:
To get us in sync with the Figma type styles. I made a pixel-by-pixel comparison [in Figma](https://www.figma.com/file/EHpOV0EI4a0aU1ezShuYMu/Title-1-comparison?node-id=1%3A5), which shows how both the before and after **web** styles compare to the "Title/X-Large" style in Figma. ~What I get from this is that the web 500 font-weight is actually more similar to the SemiBold variant than 600, but I'm curious what @coreyhunt, @laurenajong, and @nlsfds think.~ We discussed this in Slack and agreed that 600 is the way to go.

I tweaked the [typography introduction](https://sfgov-design-system-pr-66.herokuapp.com/foundations/typography/#introduction) text and links a bit and inlined the numeric font weights from our theme. One thing that doing this highlights is that we currently refer to the semibold weight as `medium`, which means you use `text-medium` to set `font-weight: 600`. This is super confusing and we should change it in the next major version and provide guidance for migrating to the new name. (IMO, for the sake of consistency, we should refer to it as `semibold` throughout.)

I've also taken this opportunity to upgrade to the [v2 Google Fonts CSS API](https://developers.google.com/fonts/docs/css2) and added `<link rel=preload>` tags to our docs HTML as [Google suggests](https://web.dev/preload-critical-assets/).

### DIFF
--- a/docs/_includes/head.njk
+++ b/docs/_includes/head.njk
@@ -7,6 +7,9 @@
 <meta charset="utf-8">
 <meta name="viewport" content="{{ meta.viewport | default('width=device-width,initial-scale=1') }}">
 <link rel="shortcut icon" href="https://sf.gov/themes/custom/sfgovpl/favicon.ico" type="image/vnd.microsoft.icon">
+{# these speed up Google Fonts requests considerably #}
+<link rel="preconnect" href="https://fonts.googleapis.com">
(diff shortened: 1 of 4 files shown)


jacksongonzales (approved):

jacksongonzales (approved):
so snappy 🚀

PR #67, opened 2021-12-30
Fix nav

jacksongonzales:
This is a refactored implementation of our navigation mechanism following [Christina's recommendations](https://docs.google.com/document/d/11xFJrbzV3TI28leQVmg_1sI2NjlvsQUylPbhCncbiis/edit) for how to improve accessibility.

The key changes are:
- Top-level nav items are now a `<details>` element that open to reveal links to the documentation. The top-level links (e.g. [/foundations](https://sfgov-design-system-pr-67.herokuapp.com/foundations/)) are pushed into the submenu as the first link.
- Tweaks to the grid layout to make the navigation links align nicely with the `.prose` content of the docs
- A new 'skip to content' link. Using the `sr-only` and `not-sr-only` utilities, the link is hidden by default but appears and is focusable when tabbed to so keyboard/screen reader users can avoid tabbing through the various nav links before picking up content
- Applying changes to decorative `<sfgov-icons>` as discussed [here](https://sfdigitalservices.slack.com/archives/CCHRVN7AB/p1640284549035800).


docs/_includes/base.njk
```diff
@@ -37,10 +37,10 @@ layout: skeleton
 <div class="responsive-container flex lg:hidden mt-60 mb-28 items-center">
   <details class="details-reset">
     <summary role="navigation" aria-label="open navigation menu">
-      <sfgov-icon class="pr-20" symbol="menu" role="img" aria-label="hamburger menu"></sfgov-icon>
+      <sfgov-icon class="pr-20" symbol="menu" aria-hidden="true"></sfgov-icon>
     </summary>
     <details-dialog class="top-0 left-0 w-full h-full fixed p-20 bg-grey-1 z-50 max-w-sm">
-      <sfgov-icon class="py-20 block mt-96 ml-8" symbol="close" role="img" aria-label="x" data-close-dialog></sfgov-icon>
+      <sfgov-icon class="py-20 block mt-96 ml-8" symbol="close" aria-hidden="true" data-close-dialog></sfgov-icon>
```
shawnbot, line 43: I think we need something with a label here so that the close button is announced by AT. It either needs `role="button"` or to be wrapped in a `<button>`, right? If we were writing a test for this, they might look like:

```feature
Given I am using a screen reader
Given my viewport has mobile width
 When I tab to the button labeled "open navigation menu"
 Then I press "Enter"
 Then the button labeled "close navigation menu" should have focus
 When I press "Enter"
 Then the button labeled "open navigation menu" should have focus
```

Where `the button labeled "{label}"` could reference any element with a button role and an `aria-label` or normalized text content equal to `{label}`.

docs/components/details.md
```diff
@@ -39,8 +39,8 @@ closed for a <code>:icon[down arrow]{symbol=chevron-down}</code> when open:
     <div class="title-xs">
       Summary
     </div>
-    <sfgov-icon symbol="chevron-right" class="open:hidden"></sfgov-icon>
-    <sfgov-icon symbol="chevron-down" class="hidden open:inline-flex"></sfgov-icon>
+    <sfgov-icon symbol="chevron-right" class="open:hidden" aria-hidden="true"></sfgov-icon>
+    <sfgov-icon symbol="chevron-down" class="hidden open:inline-flex" aria-hidden="true"></sfgov-icon>
```
shawnbot, line 43: Sorry, I assumed that we would make this change to the custom element, [here](https://github.com/SFDigitalServices/design-system/blob/354e2ff5af1ef43390bee83c4979f9478958d5e9/src/icons/index.js#L45-L49):

```diff
  connectedCallback () {
+   if (!this.hasAttribute('aria-hidden')) {
+     this.setAttribute('aria-hidden', true)
+   }
    if (!this.hasAttribute('role')) {
      this.setAttribute('role', 'img')
    }
    this.render()
  }
```

jacksongonzales: I forget, did we settle on the fact that all instances of `<sfgov-icon>` are decorative?

docs/_includes/base.njk
jacksongonzales: don't we want to entirely skip over the icon and make sure the focusable, wrapping disclosure element is providing all the contextual info? I just tested with the mac screen reader functionality and it declares whether the element is 'collapsed' or 'expanded'

shawnbot: I think this is the close button, though, right? That doesn't live in the summary, so while VoiceOver properly announces the open state, it's not announcing the close button. 

Maybe it doesn't need to because the summary is focusable? I'm not super familiar with how different AT handle details, so let's pair on this tomorrow and chat with @clutzhat 😁

jacksongonzales: ugh, I was thinking these were the chevrons for the menu items, my mistake. you're totally right, the close isn't focusable and should be 😅

docs/components/details.md
shawnbot: I think we did. If we roll with my suggestion above, then they'd be decorative by default but could be made visible by setting `aria-hidden="false"`.

docs/_includes/base.njk
```diff
@@ -36,11 +36,13 @@ layout: skeleton
 
 <div class="responsive-container flex lg:hidden mt-60 mb-28 items-center">
   <details class="details-reset">
-    <summary role="navigation" aria-label="open navigation menu">
-      <sfgov-icon class="pr-20" symbol="menu" role="img" aria-label="hamburger menu"></sfgov-icon>
+    <summary class="cursor-pointer" role="navigation" aria-label="open navigation menu">
```
shawnbot, line 39: Is `navigation` the right role here? My understanding is that [the navigation role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role) is a "landmark" that should wrap interactive elements, so maybe the role belongs on the `<details>` instead? 🤔 cc: @clutzhat

jacksongonzales: I think you're right, I'll push the change and we can loop back with Christina

shawnbot (changes requested):
I ran this through [axe DevTools](https://www.deque.com/axe/devtools/) (ping @clutzhat if you need an axe account!) and it highlighted one issue with the nav:

> ARIA role navigation must be removed when the element is made visible, as it is not allowed for the element

The `details` element with `role="navigation"` appears not to be allowed. Maybe the navigation role belongs on the `details-dialog`? Or is it then impossible to "get to" the navigation if it's hidden behind a closed `details` element? 🤔 

Using VoiceOver, there are a couple of other issues:

1. There's a lot to tab through before you get to the navigation. As @clutzhat suggested, we need a [skip navigation link](https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/G1) that comes before everything else and takes you to the page content. (Which element? Probably `<main>`?) Something like:

    ```html
    <a class="sr-only focus:not-sr-only fixed top-0 left-0">Skip to content</a>
    ```
    
    We should be able to get away with using [Tailwind's built-ins](https://v2.tailwindcss.com/docs/screen-readers) for this!  

1. It's not currently possible to navigate to the top-level sections. I think we might need to bring back the link and nudge the icon up with a negative margin or absolute positioning, e.g.

    ```jinja
    <div class="relative">
      <a
        href="{{ url }}"
        class="block title-xs no-underline mb-20"
      >
        {{ node.data.title }}
      </a>
      {% if deep and children.length %}
        <details
          class="details-reset"
          {% if page.url.startsWith(url) %} open {% endif %}
        >
          <summary
            {# i.e. "more foundations links" #}
            aria-label="more {{ node.data.title }} links"
            {# tweak positioning, add size utilities if necessary #}
            class="absolute top-0 right-0"
          >
            {# icons #}
          </summary>
          {# children list #}
        </details>
      {% endif %}
    </div>
    ```

Let's pair on these at your earliest convenience, and/or with @clutzhat if available!

jacksongonzales:
> 2. It's not currently possible to navigate to the top-level sections. I think we might need to bring back the link and nudge the icon up with a negative margin or absolute positioning, e.g.
>    ```
>    <div class="relative">
>      <a
>        href="{{ url }}"
>        class="block title-xs no-underline mb-20"
>      >
>        {{ node.data.title }}
>      </a>
>      {% if deep and children.length %}
>        <details
>          class="details-reset"
>          {% if page.url.startsWith(url) %} open {% endif %}
>        >
>          <summary
>            {# i.e. "more foundations links" #}
>            aria-label="more {{ node.data.title }} links"
>            {# tweak positioning, add size utilities if necessary #}
>            class="absolute top-0 right-0"
>          >
>            {# icons #}
>          </summary>
>          {# children list #}
>        </details>
>      {% endif %}
>    </div>
>    ```
@shawnbot I pushed a [commit](https://github.com/SFDigitalServices/design-system/pull/67/commits/3d485fc76c7467b763db5e3159bb8dc29051c87a) with a potential fix for this. Let me know what you think, in particular I'm pretty lukewarm to the fairly verbose "{{ parentPage }} Overview" language so I'm open to suggestions

jacksongonzales:
@shawnbot I think this ready to go, sry for the back and forth!

docs/_includes/main.njk
```diff
@@ -1,3 +1,3 @@
-<div class="prose">
+<div id="main-content" class="prose">
   {{ content | safe }}
 </div>
```
shawnbot, line 3: Maybe while we're in here we should turn this into a `<main>`?

```suggestion
<main id="main-content" class="prose">
  {{ content | safe }}
</main>
```

docs/usage/setup.md
```diff
@@ -0,0 +1,70 @@
+---
+title: Setup
```
shawnbot, line 2: Can we call this "Install" and rename the file to `install.md`?

```suggestion
title: Install
```

shawnbot (approved):
This looks great. Let's include it in #76 (v2.3.0)! 🚀

PR #68, opened 2021-12-30
Adds Icon documentation

coreyhunt:
This adds the documentation draft with text and screenshots.


docs/foundations/icons.md
```diff
@@ -0,0 +1,157 @@
...
+
+Interface icons, or just icons, are generally used to convey an action or intent and are not illustrative or decorative.
+
+Picture icons serve some of the same purposes as the interface icons but in a more expressive way.
+
+### Deciding on a new icon
+
+Before creating a new icon, pause and look for an icon in the SF DesSys icon library that clearly conveys your message; there’s no reason to bloat the icon library for the sake of having variety.
+
+![A calendar icon being used with the word Events](/assets/need-for-new-icon-1.png)
```
jacksongonzales, line 33: eleventy does some under-the-hood magic that means that accessing the images isn't as simple as reaching into the directory you're working in. (this is why the images aren't showing up in the deployment rn) I _think_ what we'll wanna do is shove images into a `static/` directory or find a way to put it into the `dist/` dir. maybe we can discuss next week and pair/trio on it if necessary. [here's](https://www.11ty.dev/docs/copy/) the not-so-straightforward way of getting files into the build

docs/foundations/icons.md
```diff
@@ -0,0 +1,156 @@
...
+</figure>
+
+
+### Creating a new icon
+
+Great! You’ve decided there is a need for this icon, and you know exactly what it should be. We’ve created guidelines to help your new icon feel like a cohesive member of the San Francisco Design System.
+
+#### Interface icons
+
+##### Workspace
```
jacksongonzales, line 51: I think this is the first time we've used an `h6` on the page? it's pretty tiny so I fear it won't pass visual accessibility muster, which probably means we should change it to something larger (I'm open to suggestions), _but_, it also raises whether we should implement guidelines for/against `h6` usage

docs/foundations/icons.md
```diff
@@ -0,0 +1,156 @@
...
+Before creating a new icon, pause and look for an icon in the SF DesSys icon library that clearly conveys your message; there’s no reason to bloat the icon library for the sake of having variety.
+
+<img width="250" alt="A calendar icon being used with the word Events" src="https://user-images.githubusercontent.com/957314/151463326-d263db7a-452a-4cfc-9d10-033dc09212f0.png">
+
+<img width="250" alt="A calendar icon being used with a form field" src="https://user-images.githubusercontent.com/957314/151463860-b946082b-761e-4a4c-92d8-a30b592c6ba5.png">
+
+Once you’ve decided that there is a need for a new icon, you have to decide on the best way to represent the concept. In most cases, there is a representation of that idea that already exists.
+
+<figure>
+  <img width="612" alt="Examples of delete icons" src="https://user-images.githubusercontent.com/957314/151464294-ac829556-2d12-49bb-b48d-14098d5dce13.png">
```
jacksongonzales, line 40: this image size isn't playing well with surrounding elements — I think if you change
```suggestion
  <img class="w-1/1" alt="Examples of delete icons" src="https://user-images.githubusercontent.com/957314/151464294-ac829556-2d12-49bb-b48d-14098d5dce13.png">
```

it should fix it

docs/foundations/icons.md
```diff
@@ -0,0 +1,156 @@
...
+
+It’s primarily up to the designer when to introduce an icon to their design. The designer should ask if an icon will enhance the experience of the user or create noise and confusion.
+
+### How to use
+
+###### Sizing
+
+The icons are sized at 20px and should generally not be scaled beyond that size.
+
+| <img width="250" alt="icon-size-correct" src="https://user-images.githubusercontent.com/957314/151465258-0f916e68-b94b-4183-9853-2aa8805f3ea6.png"> | <img width="250" alt="icon-size-incorrect" src="https://user-images.githubusercontent.com/957314/151465276-4ee44c10-dbfc-435f-83a0-e0ba1e342f7a.png"> |
```
jacksongonzales, line 116: the image sizes here are causing a similar problem to line 40 above. the same fix should work but would resizing the images defeat the purpose of the example?

docs/foundations/icons.md
```diff
@@ -0,0 +1,156 @@
...
+| ----------- | ----------- |
+| Yes      | No       |
+| Icon is implemented at 20px | Icon has been scaled beyond the recommended 20px |
+
+###### Margins & padding
+
+We’ve created all icons 2px padding, and that padding should not be infringed upon by other icons or text.
+
+| <img width="250" alt="icon-margin-correct" src="https://user-images.githubusercontent.com/957314/151465295-84677aab-e5f0-4c66-8935-00b3c4c7d4ac.png"> | <img width="250" alt="icon-margin-incorrect" src="https://user-images.githubusercontent.com/957314/151465312-a47a07b4-e280-4e94-9415-c2a3fb7933cc.png"> |
```
jacksongonzales, line 125: yea seems like `width=250` is too large for our page to accommodate. If you replace all instances of `width=250` with `class="w-1/1"` like above we should have at least a temporary fix. I imagine with some of these we'll want to comb through and replace images with the actual markup (e.g. text and svg's) if it's available, though.

jacksongonzales (changes requested):
this looks great! just a few suggestions/questions, lmk if you want to discuss

docs/foundations/icons.md
coreyhunt: I changed it up to remove the need for an h6. h5 may be as deep as we will want to go.

docs/foundations/icons.md
coreyhunt: Yeah, ideally, once the Do/Don't component is made, these would be replaced with live examples

coreyhunt:
@jacksongonzales , the images still look too wide even with the changes to the classes.

docs/foundations/icons.md
jacksongonzales: hm I'm still seeing `h6`'s. we 'behead' heading elements by 1, so `h1` becomes `h2` becomes `h3` and so on. although we might change this after christina made some accessibility recs, so maybe we can worry about this later

jacksongonzales (approved):

PR #72, opened 2022-02-10
feat(forms): design checkbox component

jacksongonzales:
This is majorly a WIP, but opening up so I can get feedback and pin some thoughts.
---

#### Design
* I wasn't able to find a solution to the [toggled](https://www.figma.com/file/nCDNClTAztpLol9l74QWSP/SF-Components?node-id=3861%3A4832) state without using some javascript. This creates an additional dependency for anyone who wants to use our components, so it raises a question about how flexible/rigid we are. Do we leave any `.js` implementations up to others, or do we want to create a dedicated `forms.js`? Do we wanna avoid the former entirely and shift the design to something that can be done in pure css?
* We need to clear up how the matrix of checkbox states should overlap. E.g. it's not clear right now what the hover behavior of a checkbox in error state should be. (This is also a question contingent on the above, probably 🙂 )

#### Accessibility
* I think we're gonna need to put in accessibility guidance for things like error/disabled states. From my brief research it looks like the best tool for making forms accessible is providing documentation within `aria-` states that explain why the form is in its current state, and how to proceed.
* Ask christina for her wisdom on the above 😄 
---
### Summary (wip)

This adds several component classes:

- `.form-radio` which reuses the `.form-checkbox` classes but distinguishes parent elements for clarity (this might be unnecessary, we could make a more generic class name for both)
- `.form-radio-check` which adds styles for the radio button visual and its toggle states.

The latter class required adding two new utilities: `border-6` (for creating the 'checked' state) and `rounded-full` (to make a circle).


docs/components/forms/checkbox.md
```diff
@@ -11,39 +11,45 @@ Here we will explain everything you could possibly need to know about the SF Des
 Check it out:
```
shawnbot, line 11: 😏

docs/components/forms/checkbox.md
```diff
@@ -0,0 +1,181 @@
+---
+title: Checkbox
```
shawnbot, line 2: We could link to the Figma doc here:

```suggestion
title: Checkbox
see_also:
  - title: Figma
    href: 'https://www.figma.com/file/nCDNClTAztpLol9l74QWSP/SF-Components?node-id=3861%3A4832'
```

Side note: we might also want (separately) to refactor these `see_also` keys as a more general-purpose `links` list, e.g.

```yaml
links:
  - href: 'https://figma.com/...'
    text: Figma
    rel: figma
  - href: /some/other/page
    text: Some other page
    rel: see-also
```

docs/components/forms/checkbox.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Introduction
+
+Here we will explain everything you could possibly need to know about the SF DesSys checkbox component
+
+## Checkbox component anatomy
+
+Check it out:
+
+```html
```
shawnbot, line 13: We could highlight the important bits with a regular expression. The only limitation right now is that we can't highlight across "token" boundaries (attribute names, `=`, and the quoted values are distinct tokens and wrapped in spans for syntax highlighting), so something like:

```suggestion
```html highlight="(form-checkbox[-\w]*|checkbox-id)"
```

jacksongonzales:
> I think this is looking really good! There are a couple of things that I think we need to figure out as a team:
> 
> 0. I'm still on the fence about whether we should be requiring the use of `<sfgov-icon>` for the check icon, but I don't really know how to make that decision. 🤔
> 1. Most of the `.small-checkbox-group` styles feel fussy to me for some reason. I think I'd prefer those to be expressed in the HTML rather than in CSS, but we need a better rule of thumb for which things to style implicitly (in CSS) or explicitly (in HTML). For instance: states (colors and borders) in CSS, sizing and positioning in HTML? I recognize that the check icon blurs this line.
> 2. I'm not sure if we should be committing to all of the label orientations, specifically bottom and left. We have clear use cases for above and right, and the other layouts can be achieved with utilities. See: [YAGNI principle](https://martinfowler.com/bliki/Yagni.html) 😄
> 
> What do you think?

Thanks!!

0. Hmmm, it's true that adding the icon dependency is more bulk, and it'll make the radio buttons look comparably slim. I'll try a pure CSS checkmark approach to probe how hacky it really is 😅 
1. I think that's a good rule of thumb to follow; it was a question I was fussing over too so thanks for offering a clear answer! Since we can reasonably assume the checkmark is a tight dependency, I think it's probably OK that it'll get some size/position styles.
2. Lol, good point. I'll make clear that we encourage top and right orientation

docs/components/forms/checkbox.md
shawnbot: Now that we've figured out how this should work, we need to fill out this section more explicitly.

docs/components/forms/checkbox.md
```diff
@@ -0,0 +1,170 @@
...
+      <sfgov-icon symbol="check" class="form-checkbox-check" width="16" height="16"></sfgov-icon>
+      <span>This is the label</span>
+    </label>
+  </div>
+</form>
+```
+
+## Fieldset
+
+Here's how we group together a set of chekboxes. Put in da `border-0` class on da `<fieldset>` to reset da default stylez.
```
shawnbot, line 60:
```suggestion
Two or more checkboxes should be grouped inside a `<fieldset>`. The `border-0` and `space-y-12` classes remove browser default styling and add vertical space between children.
```

docs/components/forms/checkbox.md
```diff
@@ -0,0 +1,170 @@
...
+
+```html
+<fieldset class="border-0 space-y-12">
+  <div for="foo-input" class="block title-xs">
+    Title 
+  </div>
```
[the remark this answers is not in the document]
jacksongonzales: I'm not sure; `<legend>` adds some extra default styles we'd have to reset and as far as I can tell it doesn't provide any particularly useful accessibility features. We'd likely have to add an aria-label to describe its contextual purpose either way

docs/components/forms/checkbox.md
```diff
@@ -0,0 +1,178 @@
+---
+title: Checkbox
+see_also:
+  - title: Figma
+    href: 'https://www.figma.com/file/nCDNClTAztpLol9l74QWSP/SF-Components?node-id=3861%3A4832'
+---
+
+## Introduction
+
+Here we will explain everything you could possibly need to know about the SF DesSys checkbox component
```
shawnbot, line 10: Do we have anything less placeholder-y for this? 😬

docs/components/forms/checkbox.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Checkbox component anatomy
+
+Checkboxes are made up of three constitutive parts: an input, a label, and a `<sfgov-icon>` checkmark.
```
shawnbot, line 14:
```suggestion
Checkboxes are made up of three parts: an input, a label, and a `<sfgov-icon>` checkmark.
```

docs/components/forms/checkbox.md
```diff
@@ -0,0 +1,178 @@
...
+    </div>
+    <div class="text-small text-red-4">
+      Please enter the first letter of a fruit
+    </div>
+</fieldset>
+```
+
+## Checkbox Field Alts
+
+Here are alts to the default checkbox configuration.
```
shawnbot, line 101: I think we called these "variations" elsewhere? Or was it "variants"? 🤔 

```suggestion
## Variations

```

src/css/components/forms.css
```diff
@@ -15,6 +15,115 @@
...
+    .form-checkbox-check {
+      @apply rounded-4;
+    }
+
+    .form-radio-check {
+      @apply rounded-full;
+    }
+  }
+
+  /* TODO: much of the following could be done with the `peer-` variants in tailwind 3.0.0 */
```
shawnbot, line 39: Just spitballing here, but see also: #55. I wonder if there's a tweak to the syntax that would allow us to flag dependency versions, like:

```suggestion
  /* TODO [tailwindcss@>=3]: use `peer-` variants */
```

src/theme.js
```diff
@@ -35,15 +35,17 @@ module.exports = {
     DEFAULT: '8px',
     0: '0',
     2: '2px',
-    4: '4px'
+    4: '4px',
+    full: '9999px'
```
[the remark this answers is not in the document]
jacksongonzales: good question! I just went off of tailwind's value, and am not sure why they opt for `px` over a percentage. the closest thing to an answer i could find was that it's maybe an artifact of legacy browser versions not supporting percentage with `border-radius`. now it's [fully supported](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#browser_compatibility) and they seem to behave the same, though 🤷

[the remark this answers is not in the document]
shawnbot: Cool, works for me!

shawnbot (approved):
Let's merge this!

PR #74, opened 2022-02-16
Add npm dist-tags for each published branch

shawnbot:
This PR adds another step after our npm publishing in GitHub Actions that [adds an npm dist-tag](https://docs.npmjs.com/adding-dist-tags-to-packages#publishing-a-package-with-a-dist-tag) to the released version in the form `branch.{name}`. (Note: I originally used `branch/{branch}`, but npm seems to confuse the slash with a namespace.) The idea here is that we can use the branch tags in formio-sfds and on sfgov to keep the work in sync.

There's an additional ~job~ step that should clean up the dist-tags for deleted branches. We should be able to merge this to `main` and test the workflow from start to finish without disrupting other work. You can see [here](https://github.com/SFDigitalServices/design-system/runs/5325110366?check_suite_focus=true#step:13:9) that npm published the dist-tag and confirm that it's working by visiting [unpkg.com/sfgov-design-system@branch.branch-dist-tags](https://unpkg.com/sfgov-design-system@branch.branch-dist-tags/) or running:

```sh
npm info sfgov-design-system@branch.branch-dist-tags
```

Which for me prints:

```
sfgov-design-system@0.0.0-d3a7339 | MIT | deps: none | versions: 359
... etc.
```

(The duplicate `branch` there is because this branch is literally called `branch-dist-tags` 🙃 )


.github/workflows/ci.yml
```diff
@@ -27,9 +28,21 @@ jobs:
...
+      # if the package was published, add a "branch/{branch}" dist-tag
+      # (see: <https://docs.npmjs.com/adding-dist-tags-to-packages>)
+      # so that other apps can reference the latest release from a branch
+      # with "sfgov-design-system@branch/{branch}"
+      - if: steps.publish.conclusion == 'success' && github.ref_type == 'branch'
+        env:
+          DIST_TAG: branch/${{ github.ref_name }}
+          NPM_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+        run: |
+          npm dist-tag add "$(npm info . name)@$(npm info . version)" "$DIST_TAG"
```
shawnbot, line 48:
```suggestion
          echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > $HOME/.npmrc
          npm dist-tag add "$(npm info . name)@$(npm info . version)" "$DIST_TAG"
```

jacksongonzales (approved):
looks great!

.github/workflows/ci.yml
```diff
@@ -27,9 +33,24 @@ jobs:
...
         with:
-          npm_args: --unsafe-perm --allow-same-version
+          npm_args: --unsafe-perm
           default_branch: main
         env:
           GITHUB_TOKEN: ${{ github.token }}
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+
+      - name: set BRANCH_DIST_TAG
+        if: github.ref_type == 'branch'
```
jacksongonzales, line 44: i'm an Actions newb, so apologies, but I'm guessing this line here makes sure this won't run on merge commits into `main`? assuming we don't want a branch.main along with the actual published version 😅

shawnbot: These checks just limit the step to runs when the "ref" in question is a branch, which should skip tags. I think this is necessary because `push` events include tags, and the `@primer/publish` action handles release tags automatically, so this should avoid an infinite actions loop. 🙃

.github/workflows/ci.yml
```diff
@@ -27,9 +33,24 @@ jobs:
...
+        run: echo "BRANCH_DIST_TAG=branch.${{ github.ref_name }}" >> $GITHUB_ENV
+
+      - name: npm dist-tag add
+        if: steps.publish.conclusion == 'success' && env.BRANCH_DIST_TAG
+        run: |
+          CURRENT=$(node -e 'const { name, version } = require("./package.json"); console.log(`${name}@${version}`)')
```
shawnbot, line 50: This was just the easiest way to get the name and version from `package.json` after `@primer/publish` has modified it in-place.

.github/workflows/ci.yml
```diff
@@ -1,9 +1,13 @@
 name: CI
-on: [push]
+on:
+  - push
+  - delete
```
shawnbot, line 4: Ugh, I'm looking at this now and realizing that this is going to run the whole CI job every time a ref is deleted, which will double up runs every time we merge a PR.

I think we need this in a separate job:

```yml
jobs:
  ci:
    if: github.event_name == 'push'
    # ...

  cleanup:
    if: github.event_name == 'delete'
    # ...
```

🤔

jacksongonzales: ah, yep. a separate job makes sense since the delete stuff would be a totally separate scope

PR #75, opened 2022-02-19
feat(forms): design radio component

jacksongonzales:
This adds the radio button component based on the current [figma](https://www.figma.com/file/nCDNClTAztpLol9l74QWSP/SF-Components?node-id=3861%3A3791) designs.

The implementation is similar in approach to the checkbox (#72) but adds two component classes:
- `.form-radio` which reuses the `.form-checkbox` classes but distinguishes parent elements for clarity (this might be unnecessary, we could make a more generic class name for both)
- `.form-radio-check` which adds styles for the radio button visual and its toggle states.

The latter class required adding two new utilities: `border-6` (for creating the 'checked' state) and `rounded-full` (to make a circle).

### DIFF
--- a/docs/components/forms/checkbox.md
+++ b/docs/components/forms/checkbox.md
@@ -57,7 +57,7 @@ We should probably come up with guidance about accessibility and disabled form e
 
 ## Fieldset
 
-Here's how we group together a set of chekboxes. Put in da `border-0` class on da `<fieldset>` to reset da default stylez.
+Two or more checkboxes should be grouped inside a `<fieldset>`. The `border-0` and `space-y-12` classes remove browser default styling and add vertical space between children.
 
 ```html
 <fieldset class="border-0 space-y-12">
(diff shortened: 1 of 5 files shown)


jacksongonzales:
> @jacksongonzales Visually, I think this looks great. On the CSS side could you look at styling the `<input>` directly rather than using the check icon strategy?

@shawnbot I tried that approach initially then realized we have the same problem as with the checkbox — i.e. to match the hover/focus/toggle states' styles as they're designed in [figma](https://www.figma.com/file/nCDNClTAztpLol9l74QWSP/SF-Components?node-id=3861%3A3791) right now, there needs to be a wrapping element; we can't style a child element based on parent state, so we need a pseudo-input to sit in the wrapping element (`<label>` in this case) to simulate the checked/unchecked visually. 

there's a css pseudo-class `:has()` that would solve this problem. only issue is [it isn't supported](https://caniuse.com/css-has) on any browers 😬

PR #79, opened 2022-02-28
Editing buttons docs

laurenajong:
[👀  Button docs preview](https://sfgov-design-system-pr-79.herokuapp.com/components/buttons/)


docs/components/buttons.md
```diff
@@ -1,6 +1,95 @@
...
+Buttons draw more attention than a simple text link. Because of their larger surface and color, they are visually more findable and easier to click or tap. 
+
+## Types of buttons
+
+
+### Primary
+
+<button class="btn">
+  Do something
+</button>
```
shawnbot, line 15: If you wrap this part in [fenced code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting) with the `html` language, the button will be displayed as a full-blown code example.

If that's not what you want, this will at least give it some breathing room:

```suggestion
<div class="my-28">
  <button class="btn">
    Do something
  </button>
</div>
```

docs/components/buttons.md
```diff
@@ -1,38 +1,190 @@
...
+    <sfgov-icon symbol="arrow-left"></sfgov-icon>
+    <span>Previous</span>
+  </a>
+  <a href="#" class="text-action">1</a>
+  <span>...</span>
+  <a href="#" class="text-action">8</a>
+  <a href="#" class="text-action">9</a>
+  <span>10</span>
+  <a href="#" class="text-action">11</a>
+  <a href="#" class="btn btn-secondary flex gap-8">
```
jacksongonzales, line 101: clicking these affects scroll position in a kinda jarring way; I believe doing the following will fix the page jump
```suggestion
  <a href="#/" class="btn btn-secondary flex gap-8">
    <sfgov-icon symbol="arrow-left"></sfgov-icon>
    <span>Previous</span>
  </a>
  <a href="#/" class="text-action">1</a>
  <span>...</span>
  <a href="#/" class="text-action">8</a>
  <a href="#/" class="text-action">9</a>
  <span>10</span>
  <a href="#/" class="text-action">11</a>
  <a href="#/" class="btn btn-secondary flex gap-8">
```

docs/components/buttons.md
```diff
@@ -1,38 +1,190 @@
...
+    <sfgov-icon symbol="arrow-right"></sfgov-icon>
+  </a>
+</div>
+```
+
+Icons should reinforce the meaning of the button’s text. In rare cases, an icon can be used without text. Only do this if the icon is extremely universally understood, such as an arrow or search magnifying glass. [Read more about icons](/foundations/icons/)
+
+## Appearance
+
+<img src="/static/images/button-specs.png" class="w-full">
```
jacksongonzales, line 112: I think this qualifies as a [decorative image](https://www.w3.org/WAI/tutorials/images/decorative/) since the text below explains the essence of the image, though we'll want an empty alt attribute (`alt=""`) so screen readers know to ignore the element. According to the W3C WAI guidelines, some screen readers will read the filename when an `alt` attribute is missing, which would be 🤢

jacksongonzales (commented):
do we need the [icon example](docs/static/images/icon-example.png) image since we have it in code?

docs/components/buttons.md
shawnbot: Yeah, since they don't actually do anything I'm inclined to either leave the `href` off entirely or (blech, I know) add `onclick="javascript:void(0)"`.

shawnbot: Yeah, since they don't actually do anything I'm inclined to either leave the `href` off entirely or (blech, I know) add `onclick="javascript:void(0)"`.

docs/components/buttons.md
shawnbot: Great call!
```suggestion
<img alt="" src="/static/images/button-specs.png" class="w-full">
```

jacksongonzales:
> I deleted the icon example in favor of the live markup. Let's punt on the question of reference images post-Docusaurus 😁
> 
> What should we do about the mock links? We could add a delegated `click` event listener for `[href="#"]` and prevent the event to short-circuit the page scrolling.

I like this solution for now!

jacksongonzales (approved):
🚀

PR #87, opened 2022-03-30
Extract Tailwind preset

shawnbot:
[Tailwind presets](https://v2.tailwindcss.com/docs/presets) seem like a good way to distribute this package. Usage in other projects would look like:

```js
// tailwind.config.js
module.exports = {
  presets: [
    require('sfgov-design-system/tailwind.theme')
  ]
}
```

You can then add, remove, or modify the list of core plugins, variants, and even augment the theme.

This should be part of a future minor version, since it's a new feature.


package.json
```diff
@@ -35,6 +35,9 @@
     "watch-icons": "watchy -w src/icons -- npm run build-icons",
     "start": "eleventy --serve"
   },
+  "dependencies": {
+    "tailwindcss-interaction-variants": "^5.0.0"
```
jacksongonzales, line 39: there is a duplicate declaration of this package in `devDependencies`, was this an accidental addition?

jacksongonzales: (could this be why all the webkit additions to the css package?? 🤔)

tailwind.preset.js
```diff
@@ -0,0 +1,128 @@
...
+    // TODO [^3]: nix 'focus' + 'hover'
+    textDecoration: ['focus', 'hover', 'hocus'],
+    userSelect: [],
+    verticalAlign: ['responsive'],
+    visibility: ['responsive', 'group-hocus'],
+    width: ['responsive'],
+    zIndex: ['responsive']
+  },
+
+  plugins: [
```
jacksongonzales, line 124: I noticed when trying to pull this into DAHLIA that plugins cause a bit of a hiccup by creating extra dependencies (especially the relative path'd details component). do you have any thoughts on how to handle this? installing a few dependencies isn't way out of the question, but also splitting out tokens into their own packages might solve this

jacksongonzales (approved):

tailwind.preset.js
shawnbot: Huh, we should look at this. I don't know if Tailwind is doing something special with presets, but if it's just passing through the `require()`'d plugins it shouldn't be an issue. Oh, unless this is just a matter of `tailwind-interaction-variants` being a dev dependency and `lib` not being published? 🤔

shawnbot: If that's the issue then we should just move the interaction variants to `dependencies` and make sure that `lib` gets published 😁

shawnbot: Okay, I did both of those in 344d8e915c0840540771dd604091fdf5fe6a1932 and 6fd8f7be6a211ca7c988e58d67b3cbe984ec7712. Do you mind trying again with the new canary release?

package.json
```diff
@@ -9,6 +9,7 @@
   "files": [
     "dist",
     "docs",
+    "lib/tailwind",
     "src",
     "tailwind.config.js"
```
jacksongonzales, line 14: i think the last one we need here is `tailwind.preset.js` since we require it in the config

shawnbot: Ah, of course! Okay, 8e771be publishes this should work? 🤞

jacksongonzales (approved):
looking good when I bring it into DAHLIA! thanks for making those changes 🙂

PR #89, opened 2022-04-27
Add Heroku review apps workflow

shawnbot:
I'd like to test out [this action](https://github.com/fastruby/manage-heroku-review-app) to bring back Heroku review apps. 🤞


jacksongonzales (approved):
i know we recently had some travails over stale review apps hanging around and cluttering our limit — is that something we handled on the heroku end? either way 🥳 for a solution

PR #94, opened 2022-05-19
Set up Docusaurus!

shawnbot:
We're using [Docusaurus](https://docusaurus.io) because it's awesome and has most of the problems we're trying to solved figured out. See DESSYS-204 for the gory details.

This move also prompted the creation of a [monorepo] that allows us to separate documentation (website) dependencies from the "core" `sfgov-design-system` package, and prepares us for the abstraction of other system parts: design tokens, React components, etc.

- [x] Merge #96 into this branch for Codespaces support
- [x] Get baseline linting for .md,.mdx files #100 
- [x] Custom Playground component + style encapsulation in #102 
- [x] #103 
- [x] #104 
- [x] #105 and #106 😬 

This will release `sfgov-design-system` package version `2.5.0` 🥳 

[monorepo]: https://monorepo.tools/

PR #96, opened 2022-05-26
Set up GitHub Codespaces

shawnbot:
This builds on #94 and attempts to set us up to use [GitHub Codespaces] for working on docs in the ☁️ . Here's what appears to be the quickest way to test it:

### Create a new codespace
If you want to test out the whole flow, click on the <kbd>&lt;&gt; Code</kbd> button in the top right of this PR and click the "Create codespace on [...]" button:

image

That should start up a new container running the code in this PR and run `npm install` automatically.

### Start the server
Once you're in the codespace, you should be able to just press <kbd>F5</kbd> to start ("launch", in VS Code parlance) Docusaurus with the debugger. The status bar at the bottom will report "Building..." until it's ready:

image

The site should either open in a new tab/window automatically and/or you'll see a toast in the bottom right prompting you to open it in your browser. If you miss this (it disappears after about a minute) you can navigate to the "Ports" tab and hover over the "Local Address" field in the first row, then click the globe icon to open it:

image

#### Without debugging
If you don't need to debug, you can instead run the "start" task one of several ways:

- <kbd>Cmd+Shift+P</kbd> then type "run task"

    image

- Navigate to "Terminal" → "Run task" from the top left menu:

    image

Then, in the prompt, select "start":

image

### Problems
- Live reload doesn't appear to work (or it's just _incredibly_ slow). I've had to refresh pages that I'm editing a couple of times to see the change live.
- I've noticed a lot of 504 Gateway Timeout errors when viewing the site running in the codespace. Maybe a more powerful machine would be better?

[github codespaces]: https://github.com/features/codespaces


.editorconfig
```diff
@@ -0,0 +1,15 @@
+# http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
```
shawnbot, line 9: I grabbed this from the Docusaurus repo. This file is for editors other than VS Code, e.g. Sublime, and mimic the `editor.tabSize` and `editor.useTabStops` settings in `.vscode/settings.json`. We might want to just delete the rest of the lines in this file since I haven't tested them and they probably aren't important ([YAGNI](https://martinfowler.com/bliki/Yagni.html)).

package.json
```diff
@@ -24,6 +25,7 @@
     "eslint-plugin-unicorn": "^42.0.0",
     "shellcheck": "^1.1.0",
     "stylelint": "^14.8.4",
-    "stylelint-config-standard": "^25.0.0"
+    "stylelint-config-standard": "^25.0.0",
+    "wireit": "^0.4.3"
```
shawnbot, line 29: [Wireit](https://github.com/google/wireit#readme) is awesome. I'm using it here to set up clearer dependencies between our build steps, primarily so that the website only starts after the `sfgov-design-system` package builds.

packages/sfgov-design-system/package.json
```diff
@@ -14,24 +14,101 @@
...
+    },
+    "lint": {
+      "dependencies": [
+        "lint-css",
+        "lint-js",
+        "lint-scripts"
+      ]
+    },
+    "lint-js": {
+      "command": "eslint --color --cache --cache-location=.eslintcache src lib '*.js' 'scripts/*.js'",
```
shawnbot, line 87: As recommended by [this recipe](https://github.com/google/wireit#eslint)

coreyhunt (approved):
Not much to add beyond it works great. I'm glad you were also able to simplify the process

PR #98, opened 2022-06-01
Delete Heroku actions workflow

shawnbot:
We don't need it anymore since [Incident 2413](https://status.heroku.com/incidents/2413) was resolved. 🎉


jacksongonzales (approved):
🥳

PR #102, opened 2022-07-28
Docusaurus: add custom Playground component

shawnbot:
This swizzles the Playground component and styles it to match [Figma](https://www.figma.com/file/089UadxdkLJ0xP5vAWj5V2/DesSys-Site-Design?node-id=1844%3A19491):

image

This component supports two additional values in the code block "meta" string (anything following the <code>```html</code> header):
- `closed` will collapse the code sample by default, which might be useful for more lengthy markup examples
- `class="..."` will pass the classnames along to the preview div, e.g. `class="bg-blue-dark"` to fill the whole preview area with a dark blue background

### Style encapsulation
Another issue I'm attempting to solve here is _style encapsulation_: ensuring that Docusaurus's global CSS doesn't affect our example renderings in both code playgrounds and do/don't visuals. The solution I've landed on for now is using the [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) via [react-shadow](https://github.com/Wildhoney/ReactShadow) and linking to the [published `sfds.css`](https://unpkg.com/sfgov-design-system/dist/css/sfds.css) in a scoped `<style>` tag. A couple of places where you can see this in action:

- The [link button example](https://sfgov-design-system-pr-94.herokuapp.com/components/buttons/#link-buttons) inherits hover styles from Docusaurus' CSS that made text illegible. [Fixed!](https://sfgov-design-system-pr-102.herokuapp.com/components/buttons/#link-buttons)
- The [keyboard examples](https://sfgov-design-system-pr-94.herokuapp.com/components/kbd) all inherited `kbd` styles from Docusaurus. [Fixed!](https://sfgov-design-system-pr-102.herokuapp.com/components/kbd)


website/src/components/EncapsulatedStyleRoot.jsx
```diff
@@ -0,0 +1,16 @@
...
+/**
+ */
+export default function EncapsulatedStyleRoot ({ children, ...props }) {
+  return (
+    <root.div {...props}>
+      <style type='text/css'>@import url(https://unpkg.com/sfgov-design-system@{sfdsVersion}/dist/css/sfds.css);</style>
+      {children}
+    </root.div>
+  )
+}
```
shawnbot, line 16: There's got to be a better way to do this, but it works for now. [ReactShadow](https://github.com/Wildhoney/ReactShadow) does accept a `styleSheets` prop, which is an array of CSSStyleSheet references. But I don't feel like wrestling with webpack right now in order to make this work, so every encapsulated style root gets its own `<style>@import url()</style>` linking to the published `sfds.css` with the same version that we declare in `website/package.json`.

Ideally, we would do it with inline CSS: e.g. using [string-loader] to `import` the CSS as a string, and either dumping that into a global stylesheet then referencing it (e.g. with a `useEffect()` hook), or just inlining the whole shebang:

```jsx
import sfdsStyle from 'sfgov-design-system/dist/css/sfds.css'

export default function EncapsulatedStyleRoot ({ children, ...props }) {
  return (
    <root.div {...props}>
      <style type='text/css'>{sfdsStyle}</style>
      {children}
    </root.div>
  )
}
```

.eslintrc.js
```diff
@@ -13,7 +13,11 @@ module.exports = {
     }],
     'no-trailing-spaces': ['warn', {
       ignoreComments: true
-    }]
+    }],
+    'padding-line-between-statements': [
+      'warn',
+      { blankLine: 'always', prev: '*', next: 'function' }
```
shawnbot, line 19: I've found myself liking [this eslint rule](https://eslint.org/docs/latest/rules/padding-line-between-statements): it allows us to suggest a blank newline preceding function declarations.

PR #104, opened 2022-08-22
Move redirects to nginx config

shawnbot:
This PR moves the redirects out of Docusaurus (where they're done client-side) and into our [nginx config](https://github.com/SFDigitalServices/design-system/blob/390ce52f7cbc13d5a372bdffaa72cc66bb3f68f5/config/nginx.conf.erb#L32-L49), recently broken out in #103.


jacksongonzales (approved):
I was too scared to touch nginx redirects so thx for taking the plunge 😅

PR #105, opened 2022-08-22
Fix icons

shawnbot:
This is a fix for DESSYS-231: issues with icon rendering on the docs site. I took this as an opportunity to:

1. Clean up the `<sfgov-icon>` styling situation, removing the duplicate CSS injection bits from the JS and adding an `icons.css` bundle, which is now part of `sfds.css`
2. Improve client-side rendering performance of our encapsulated (shadow root) examples with inlined CSS rather than the `@import`-ed unpkg.com URLs

    🤘  One huge benefit of this change is that we're no longer linking to any unpkg.com URLs in the docs site, which means that we can make changes to the system and see them in the docs of every review app rather than having to wait for a release

4. Importing the system's `font.css` bundle to get the official font stack


packages/sfgov-design-system/package.json
```diff
@@ -63,8 +64,9 @@
     "build-js": {
       "command": "rollup -c",
       "files": [
+        "package.json",
         "src/js/**/*.js",
-        "*.config.js"
+        "*.js"
```
shawnbot, line 69: I noticed that wireit was skipping JS builds even though some files had changed, so I widened the scope of files that it considers as changed candidates.

packages/sfgov-design-system/package.json
```diff
@@ -79,6 +81,10 @@
     },
     "build-manifest": {
       "command": "../../scripts/manifest.js 'dist/**/*.{css,*js,svg}' > manifest.json",
+      "dependencies": [
+        "build-css",
+        "build-js"
+      ],
```
shawnbot, line 87: I think this is why the manifest wasn't working properly in the docs sometimes: wireit needs to know that those other build tasks are dependencies of the manifest to queue them up in the right order.

website/src/components/EncapsulatedStyleRoot.jsx
```diff
@@ -1,16 +1,21 @@
 import React from 'react'
 import root from 'react-shadow'
+import useIsBrowser from '@docusaurus/useIsBrowser'
+// eslint-disable-next-line import/no-webpack-loader-syntax
+import sfdsStyles from '!!raw-loader!sfgov-design-system/dist/css/sfds.css'
```
shawnbot, line 5: I prefer not to muck with the webpack config 🤢

website/src/sfds.js
```diff
@@ -0,0 +1,5 @@
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment'
+
+if (ExecutionEnvironment.canUseDOM) {
+  require('sfgov-design-system')
+}
```
shawnbot, line 5: Moving this into the client module bundle ensures that it executes before any of the MDX content is rendered, which appears to fix the icons not displaying.

website/src/css/custom.css
```diff
@@ -6,8 +6,9 @@
  * work well for content-centric websites.
  */
 
+@import 'sfgov-design-system/dist/css/sfds.css';
+@import 'sfgov-design-system/dist/css/fonts.css';
 @import 'masonry-styles.css';
-@import url('https://fonts.googleapis.com/css2?family=Rubik:wght@300..900&family=Roboto+Mono&display=swap');
```
shawnbot, line 10: The docs should use our font stack so that changes to the system will apply here automatically ✨ 

https://github.com/SFDigitalServices/design-system/blob/6a8f3b7ed3ce3afec8992534811ceac9ba0fa7ba/packages/sfgov-design-system/src/css/fonts.css#L1

website/src/sfds.js
shawnbot: Also note that if we remove the `ExecutionEnvironment.canUseDOM` check, the build fails because _something_ in our browser bundle tries to call a `document` method in Docusaurus' server-side rendering ("execution") environment.

PR #106, opened 2022-08-24
Fix <sfgov-icon> in Docusaurus/React

shawnbot:
My first attempt at fixing this was to use [memoization](https://nickymeuleman.netlify.app/blog/react-memo) as a workaround to [skip child rendering](https://javascript.plainenglish.io/can-usememo-skip-a-child-render-94e61f5ad981), but I didn't have much luck with that approach. Instead, I went with a [useLayoutEffect](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) hook that forcibly re-renders the icon.

The first time I viewed [the home page](https://sfgov-design-system-pr-106.herokuapp.com/) it glitched a bit, but I can't get it to do that again, even in an incognito window. So I think it's working correctly now!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "shellcheck": "^1.1.0",
         "stylelint": "^14.8.4",
         "stylelint-config-standard": "^25.0.0",
-        "wireit": "^0.4.3"
+        "wireit": "^0.7.1"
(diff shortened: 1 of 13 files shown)


coreyhunt (approved):
Works in the test link

PR #107, opened 2022-08-25
Set up Storybook

shawnbot:
My thinking here was to set up Storybook in its own directory separately from Docusaurus (in `website`) and the design system package, which allows us to cordon off its wild tree of `@storybook/blah-blah-blah` dependencies. My plan was to then have the storybook config (`storybook/.storybook/main.js` load stories from all of the other packages and present them in a single UI. (The alternative [composition](https://storybook.js.org/docs/react/sharing/storybook-composition) setup would involve each package having its own `.storybook` directory and Storybook-related dependencies, which seems like a maintenance nightmare.)

This mostly works, and with a bit of hacking I was able to shoehorn the static Storybook build into the directory generated by Docusuarus:

* [Docusaurus](https://sfgov-design-system-pr-107.herokuapp.com/) built in `website/public`
* [Storybook](https://sfgov-design-system-pr-107.herokuapp.com/storybook/) built in `storybook/public` and copied to `website/public/storybook` post-build

Using `NODE_ENV === 'production'` as a signal for whether we're building both Docusaurus and Storybook on Heroku (which makes it simple to test locally, i.e.`NODE_ENV=production npm run build`) I was able to get them both linking to one another: Docusaurus has a Storybook link in the top right of the main nav, and Storybook links back to Docusaurus from the "San Francisco DESIGN SYSTEM" logo (imported directly from the website's SVG!). They both open in new tabs, but I'm okay with that as a first pass.

So yeah, it's ready to start putting stories into!


.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
shawnbot, line 1: `npx storybook init` flagged this and prompted me to add it automatically. It doesn't appear to impact the build in CI or on Heroku, but [npm warns against using it](https://docs.npmjs.com/cli/v8/using-npm/config#legacy-peer-deps), so 🤷

package.json
```diff
@@ -11,8 +12,10 @@
   },
   "scripts": {
     "build": "lerna run build",
-    "start": "npm start -w website",
     "lint": "lerna run lint",
+    "postbuild": "cp -rf storybook/public/ website/public/storybook/",
+    "start": "npm start -w website",
+    "storybook": "npm start -w storybook",
```
shawnbot, line 18: It would be rad if we could run these in parallel for development, or even bind them together into a single dev server... 🤔

website/docusaurus.config.js
```diff
@@ -44,7 +56,8 @@ module.exports = {
       'classic',
       /** @type {import('@docusaurus/preset-classic').Options} */
       ({
-        debug: true,
+        // uncomment this if you need to debug in production (when deployed)
+        // debug: true,
```
shawnbot, line 60: Docusaurus [recommends](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-debug) leaving this undefined to disable the [debug pages](https://docusaurus.io/__docusaurus/debug) in production (in our case, Heroku). I don't know how useful they are for us, but presumably it's faster to build without them, and we can enable them by uncommenting this line if we need to file an issue with Docusaurus.

website/package.json
```diff
@@ -2,8 +2,9 @@
   "name": "design-system.sf.gov",
   "private": true,
   "scripts": {
-    "build": "docusaurus build --out-dir public",
+    "build": "wireit",
     "clear": "docusaurus clear",
+    "link-storybook": "ln -sf static/storybook ../storybook/public",
```
shawnbot, line 7: This symlinks the storybook directory into the one built by Docusaurus. I actually don't know how useful this is since it really only makes it easier (possible?) to run `docusaurus serve` and see what it would look like in production. 🤔

package.json
```diff
@@ -3,32 +3,45 @@
...
   "workspaces": [
     "packages/*",
+    "storybook",
     "website"
   ],
   "engines": {
-    "node": ">=14.x",
-    "npm": ">=8.x"
+    "node": "16.x",
+    "npm": "8.x"
```
shawnbot, line 11: I'm not sure what's going on, but my hunch is that some failed builds were because of Heroku running Node.js 18.x. For now I think it makes sense to lock this down to 16.x so that we're running the same major version everywhere.

coreyhunt (approved):

PR #109, opened 2022-08-29
Clean up dev dependencies

shawnbot:
This removes the unused `lib/git.js` and 27 (!) development dependencies from the `sfgov-design-system` package.


jacksongonzales (approved):
🥵 🎉

PR #115, opened 2022-09-01
📝 Add back Documentation docs page

shawnbot:
A couple of post-Docusaurus tasks in this PR:

- Add (back) a [documentation](https://sfgov-design-system-pr-115.herokuapp.com/develop/documentation) page, which:
    - explains that we use Docusaurus
    - documents how the code playgrounds work
    - documents and demos our other docs-specific components, `DoDont` and `InlineSwatch`
    - demos how most common Markdown ("prose") elements look in one place (a "kitchen sink", if you will)
    - lists all of the admonition types, which should make it easy to replace them with our own alert component
- Add a redirect rule from `/usage/docs` (the old URL) to the new one (`/develop/documentation`)
- Add an MDXComponents export where we can both override Markdown components (I started with more consistent styling of `inlineCode` and fixing the fonts in `h2`) and add globals ([scope](https://docusaurus.io/docs/next/markdown-features/react#mdx-component-scope)) to MDX context so that we don't have to explicitly `import` docs components or `tokens` on each page
- A new tokens library page #114


coreyhunt (approved):

website/docs/libraries/tokens.mdx
```diff
@@ -0,0 +1,36 @@
...
+<TokenList field='fontFamily' />
+
+### Font size
+<TokenList field='fontSize' value={FontSizeValue} />
+
+### Font weight
+<TokenList field='fontWeight' />
+
+### Letter spacing
+<TokenList field='letterSpacing' />
```
shawnbot, line 28:
```suggestion
### Font family

<TokenList field='fontFamily' />

### Font size

<TokenList field='fontSize' value={FontSizeValue} />

### Font weight

<TokenList field='fontWeight' />

### Letter spacing

<TokenList field='letterSpacing' />
```

website/docs/develop/documentation.mdx
```diff
@@ -0,0 +1,201 @@
...
+}
+
+* {
+  box-sizing: border-box;
+}
+```
+
+```js
+// JavaScript
+;(async () => {
```
jacksongonzales, line 141: should there be a semi here? or is this some esoteric js i don't know about?

shawnbot: It's [esoteric](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#automatic_semicolon_insertion). 😬

shawnbot: Sorry, in the spirit of _not_ being esoteric: it's to do with automatic semicolon insertion rules and, more specifically, the  `(` as a _continuation character_ that implies a function call on the previous token/identifier if there _isn't_ an explicit semicolon. This is the classic example:

```js
const foo = require('bar')
/* this comment is removed by the interpreter */
(async () => {
  await foo()
})()
```

which interpreters will turn into:

```js
const foo = require('bar')(async () => { await foo() })()
```

which tries to assign `foo` by evaluating `require('bar')(...)`, which... pretty much never works. 🙃 :neckbeard: 🤘

shawnbot: Sorry, this is dumb. Let's just rewrite the JS example so as not to raise eyebrows unnecessarily.

website/src/components/TokenList.jsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react'
```
jacksongonzales, line 1: i think this looks great, but I wonder if we'd want to align the design of this component with `UtilityListItem`. maybe they're doing different enough things, but I'd be curious corey's thoughts

jacksongonzales: (this would def be a separate ticket)

shawnbot: Thanks for flagging. I've filed a ticket for it: DESSYS-241

jacksongonzales (approved):
🎉

PR #117, opened 2022-10-05
Stub out component stories with Figma links

shawnbot:
This addresses both DESSYS-245 (skeleton) and DESSYS-239 (Figma designs in Storybook). Check it out:

[👀  Storybook](https://sfgov-design-system-pr-117.herokuapp.com/storybook/)

I've attempted to stub out all of the components in [this Figma file](https://www.figma.com/file/nCDNClTAztpLol9l74QWSP/Design-System-Components?node-id=5895%3A3405), grouped by page when appropriate and with form-specific components nested separately. I took the liberty of implementing a couple of stories just to get a feel for the workflow:

- All of the [Alerts & Banners](https://sfgov-design-system-pr-117.herokuapp.com/storybook/?path=/story/components-alerts-banners--banner)
- All of our [Buttons](https://sfgov-design-system-pr-117.herokuapp.com/storybook/?path=/story/components-buttons--primary-button) with a table layout showing the matrix of (inline | block) &times; (rest | hover/focus)
- A few simple [Containers](https://sfgov-design-system-pr-117.herokuapp.com/storybook/?path=/story/components-containers--disclaimer)
- An [Icon component](https://sfgov-design-system-pr-117.herokuapp.com/storybook/?path=/story/components-icons--icon)
- [Keyboard instructions](https://sfgov-design-system-pr-117.herokuapp.com/storybook/?path=/story/components-keyboard--kbd)
- [Tiles](https://sfgov-design-system-pr-117.herokuapp.com/storybook/?path=/story/components-tiles--event-tile) were a lot of fun

### Linting challenges
One big issue is JavaScript linting (with ESLint), and it's a weird two-fer:

* In VS Code, ESLint doesn't appear to respect either the React plugin and/or Babel config, because the linter won't parse JSX files — it bails on each file with the (sadly, familiar) error:
    > Parsing error: This experimental syntax requires enabling one of the following parser plugin(s): "jsx", "flow", "typescript". (42:70)

    What this means is that we don't get proper linting or auto-formatting in Storybook sources, which stiiiinks. I think it might come down where ESLint is running in VS Code, because I get this error when I run `eslint  storybook/stories/**/*.jsx` from the project root but _not_ when running it from within the `storybook` directory. It might mean having to set `root: true` in `storybook/.eslintrc.js` and configuring it manually there rather than inheriting the root config.

* In GitHub Actions, `npm run lint` from the project root works just fine (!!), but ESLint flags the `sfgov-design-system` imports in Storybook with `import/no-unresolved` errors. I've worked around this by downgrading that rule to a warning. This is super confusing because `sfgov-design-system` _does_ get symlinked to `/node_modules` by npm in both environments and Storybook builds just fine. This seems like less of an issue than what's going on in VS Code.

Together, these issues make it hard to catch issues before you push to GitHub. ☹️


storybook/.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     'plugin:react/recommended'
   ],
   rules: {
-    'react/no-unescaped-entities': ['warn']
+    'react/no-unescaped-entities': ['warn'],
+    'react/prop-types': ['off']
```
shawnbot, line 14: @jacksongonzales I haven't added propTypes to any of the Storybook components (internal or story) yet. Do you have an opinion on linting for prop-types validation on all props, add propTypes for runtime validation, or... use TypeScript? 😬

storybook/.eslintrc.js
```diff
@@ -3,13 +3,16 @@ module.exports = {
...
   ],
   extends: [
     'plugin:sfgov/recommended',
     'plugin:react/recommended'
   ],
   rules: {
-    'react/no-unescaped-entities': ['warn']
+    'react/no-unescaped-entities': ['warn'],
+    'react/prop-types': ['off'],
+    'import/no-unresolved': ['warn']
```
shawnbot, line 16: I've had to switch this to a warning because some mysterious difference in the Actions environment (from my local VS Code, anyway) is preventing ESLint from seeing the `sfgov-design-system` package, even though Storybook builds it just fine.

storybook/.storybook/preview.jsx
```diff
@@ -1,37 +1,33 @@
...
-      order: [
-        '*',
-        'Components',
-        'Experiments',
-        'Internal',
-        'Docusaurus'
-      ]
-    }
+  design: {
+    name: 'Figma',
```
shawnbot, line 13: `name` controls the label of the tab in the Storybook control panel:

image

storybook/package.json
```diff
@@ -11,6 +11,7 @@
     "build": "build-storybook -o public"
   },
   "dependencies": {
+    "clsx": "^1.1.0",
     "sfgov-design-system": "2.5.0"
```
shawnbot, line 15: I broke these out as runtime dependencies, but they don't really have to be. It's possible that moving them into `devDependencies` would solve the `import/no-unresolved` issue I mentioned above? 🤔

storybook/stories/Components/IconInfoLockups.stories.jsx
```diff
@@ -0,0 +1,27 @@
+// import React from 'react'
+
+/** @type {import('@storybook/addons').StoryContext} */
+export default {
+  title: 'Components / Icon + Information Lockups',
```
shawnbot, line 5: The naming for this one doesn't really jive with the rest of the component titles. Maybe "Info blocks" is a better name? @coreyhunt

coreyhunt (approved):

PR #119, opened 2022-10-14
Components index + Storybook source link addon

shawnbot:
We've discussed in DESSYS-234 how to link Storybook and Docusaurus. This PR proposes a single source of truth for our official [components list](https://github.com/SFDigitalServices/design-system/blob/feature/components-index/website/components.js), which in turn powers links to and between the documentation and Storybook.

### Storybook
In #117 we added the [sourcelink](https://storybook.js.org/addons/storybook-source-link/) Storybook addon, which is intended to link to the _source_ code for a story via its `sourceLink` and `sourceLinkPrefix` [parameters][]. This adds a (git repo) button to the story toolbar that turns active (blue) when the `sourceLink` parameter is present. It's a bit of a hack (in that I've overloaded the meaning of "source link"), but it's simple and requires zero custom code.

image image

This PR includes a custom Storybook addon that populate's a story's `sourceLink` [parameters] from entries in the component index. Currently matches are made when the story's id (e.g. `components-button--primary`) matches the component metadata's `story.path` glob pattern (e.g. `components-button**` for all buttons), but we could be more explicit about it. If this approach turns out to be difficult or unwieldy in any way, I'm happy to nix the addon and move to using inline `sourceLink` hrefs in our code.

### Docusaurus
On the docs side, I've set up an [All components](https://sfgov-design-system-pr-119.herokuapp.com/components/all/) page that lists the components and links to both their documentation and storybook pages:

image

As you can see from that page, I have not yet:
1. Proposed or defined any of the values for the component "Status" column
2. Moved the Figma links into the component index

⚠️ **There are other ways to do this!** For instance, each component's documentation page could put the component metadata in its front matter, and the "All components" page would summarize those pages by extracting their front matter.; and/or we could link to the docs from story [parameters] and parse those out from Docusaurus. But both of those approaches are trickier to build and maintain, I think, and having a centralized list lets us do things like ensure that all of the URLs exist in the docs and stories get generated at build time, so we could (theoretically) fail a CI build if we break a component link.

[parameters]: https://storybook.js.org/docs/react/writing-stories/parameters


storybook/.storybook/preview.jsx
```diff
@@ -13,6 +13,7 @@ addParameters({
     name: 'Figma',
```
jacksongonzales, line 13: not sure if we fully committed to ditching the Figma panel, but if removing some of the figma links in components is a stride toward that, I'm personally fine with scrapping it all (until/unless we find a good use for it)

joni-f-g (approved):

website/components.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * @type {ComponentMeta[]}
+ */
+const components = [
+  {
+    name: 'Button',
+    href: '/components/buttons/',
+    story: {
+      path: 'components-buttons*'
```
jacksongonzales, line 28: the path I get for each of the stories is prepended with a `/story/components-{foo}` 

I think if we just update the baseUrl in the `StorybookLink` component to include the `/story/` in the URL these links will work

shawnbot: Thanks, I forgot to update that component! Will do.

storybook/.storybook/preview.jsx
jacksongonzales: oops, those are indeed working in the table! my b

PR #132, opened 2022-11-09
add story and initial component for spotlights

jacksongonzales:
A few notes/todos:

- Should we look into being more strict about types, e.g. accepting React children for the spotlight body?
- Are there other background colors spotlights can have?
- There's some odd behavior with images in the `primary: false`, `isTitleFirst: true` configuration when viewing in mobile. The background peeks out from beneath the image. I experimented with using `background-image` instead but that felt equally as clunky. I'll look into this some more, but I'm vexed 😕 
Screen Shot 2022-11-09 at 2 58 10 PM


storybook/src/Title.jsx
```diff
@@ -0,0 +1,10 @@
+import tw from 'tailwind-styled-components'
+
+const Title = tw.p`
```
shawnbot, line 3: Nit: can we call this `PageTitle` to be more specific?

storybook/stories/components/Spotlight.stories.jsx
```diff
@@ -0,0 +1,124 @@
...
+            <Content title={title} body={body} buttonContent={buttonContent} primary={primary} />
+          </ContentComponent>
+          <ContentComponent $as='img' src={image} isTitleFirst={isTitleFirst} {...contentProps} />
+        </>
+        : <>
+          <ContentComponent $as='img' src={image} isTitleFirst={isTitleFirst} {...contentProps} />
+          <ContentComponent isInverseTitlePanel={primary ? '' : 'true'}>
+            <Content title={title} body={body} buttonContent={buttonContent} primary={primary} />
+          </ContentComponent>
+        </>
```
shawnbot, line 120: I think that we should avoid reordering elements in the DOM if we can. The `flex-direction` utilities could do the heavy lifting of conditionally reversing the content order and simplify the code:

```js
// fka SectionComponent
const SpotlightContainer = tw.div`
  /* ... */
  ${p => p.$reverse ? 'lg:flex-row-reverse' : ''}
`
```

Then, later:

```js
return (
  <SpotlightContainer $reverse={!isTitleFirst}>
    <SpotlightText {...textProps} />
    <SpotlightImage {...imageProps} />
  </SectionComponent>
)
```

If you go this route I would also suggest changing the variable name to `isImageFirst` so that you don't have to negate it. IMO the default should be text first (`isImageFirst = false`), but the final call is @coreyhunt's.

storybook/src/Title.jsx
shawnbot: Oh also, this should probably be an `<h2>` by default rather than `<p>` 😁

PR #135, opened 2022-11-17
add section and tile components for page stories

jacksongonzales:
Adds:
- News, Feature, and Content tiles
- Section and section title components
- Two stories:
   - One for tiles
   - Another for Page Sections (News and Services sections)

TBD:
- Where do we use Event Tiles? I wasn't able to find any use of them on sf.gov
- Should we rename the SectionTitle to something more generic?


coreyhunt:
@jacksongonzales I agree with your intuition of making SectionTitle more generic.
"Title" would probably be adequate.

coreyhunt (approved):

PR #136, opened 2022-11-19
Stub out React components

shawnbot:
This PR stubs out a new npm package, `@sfgov/react` (and another one for the icons, `@sfgov/icons`). Technical details:

- All components and support code written in (and [linted with][eslintrc]) TypeScript
- ESM and CJS builds with Rollup + Babel. This setup gives us the ability to compile down TypeScript, use [Babel macros][] (in the future), and selectively bundle dependencies. Currently, we bundle every dependency that _isn't_ listed in the package's `peerDependencies`.
- [Stitches] used for React CSS, which has a lot of nice benefits:
    - Zero peer dependencies, which dramatically simplifies installation and clarifies total bundle size
    - Slim runtime: we use [createStitches()](https://stitches.dev/docs/api#createstitches) and ship our own themed implementations of the entire [Stitches API]
    - No need for wrapping our components in a `<ThemeProvider>` because they're themed out of the box
    - A much simpler API that's similar to styled-components (minus the tagged template literals) but incorporates themed CSS props a la Theme UI
    - First-class [variant support][variants] that's declarative and super flexible
    - More performant TypeScript types and improved DX
- [React 18] across the board (as a peer dependency and upgraded in Docusaurus + Storybook), because Theme UI requires it

Includes:
- [x] Refactor from Theme UI to Stitches in #139 
- [x] Initial React icons implementation in #140 and improvement in #142 
- [x] Lerna publish CI workflow in #143 
- [x] Server-side and global CSS helpers in #144 
- [x] Page component refactor in #146 

[react 18]: https://reactjs.org/blog/2022/03/29/react-v18.html
[eslintrc]: https://github.com/SFDigitalServices/design-system/blob/feature/react-package/packages/react/.eslintrc.js
[babel macros]: https://jakerunzer.com/babel-macros
[stitches]: https://stitches.dev/
[stitches api]: https://stitches.dev/docs/api
[variants]: https://stitches.dev/docs/variants


.github/workflows/ci.yml
```diff
@@ -18,17 +20,27 @@ jobs:
...
+          git remote set-url origin "https://x-access-token:${GITHUB_TOKEN}@github.com/$GITHUB_REPOSITORY"
         env:
-          GITHUB_TOKEN: ${{ github.token }}
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - run: echo "NPM_AUTH_TOKEN=${{ secrets.NPM_AUTH_TOKEN }}" >> $GITHUB_ENV
+      - name: Publish latest (from 'main' branch)
+        if: github.ref == 'refs/heads/main'
+        run: lerna publish from-package --yes
```
shawnbot, line 38: I'm looking at [this example](https://github.com/jonwa/lerna-release-workflow)'s [workflow](https://github.com/jonwa/lerna-release-workflow/blob/master/.github/workflows/release.yml#L35-L38) and corresponding [package scripts](https://github.com/jonwa/lerna-release-workflow/blob/master/package.json#L16) for inspiration, and it seems like we might need to tweak it if we want new versions generated from [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/):

```suggestion
        run: |
          lerna version --yes --conventional-commits --create-release github --message 'chore(release): publish'
          lerna publish from-package --yes
```

I honestly don't know what should or will happen the first time this runs, so we should probably test it once locally (either without `--yes` or with `--dry-run`) before merging. Ideally, running it locally would publish the versions currently listed in git and bump each package's version according to the conventional commits in the diff, then merging to `main` would be a no-op. A slightly worse outcome would be a CI failure; the worst, IMO, would be if it bumps the versions _again_.

.github/workflows/ci.yml
```diff
@@ -18,17 +20,27 @@ jobs:
...
+        id: publishMain
+      - name: Publish prerelease (pull requests)
+        if: startsWith(github.ref, 'refs/heads/release-')
+        run: lerna publish prerelease --dist-tag next --yes
+        id: publishPre
```
shawnbot, line 43: I actually don't think there _is_ a prerelease step if we use conventional commits: every push publishes canary versions for only the affected packages, and merging to `main` derives the published version from the commit messages.

shawnbot (commented):
A few things I'd like to do before merging this:

1. Get it deploying. I tried starting a new review app and the build failed 😱 
2. Update the React docs on the website — maybe just include the react package's README?
3. Figure out the Lerna version/publish workflow so that we can be more confident about what happens when we merge.

jacksongonzales (approved):

PR #139, opened 2022-12-02
Refactor with Stitches

shawnbot:
This PR replaces Theme UI (and the inherited `@emotion/react` peer dependency) with [Stitches]. There are some big upshots to this:

- Zero peer dependencies, which dramatically simplifies installation and clarifies total bundle size
- Slim runtime: we use [createStitches()](https://stitches.dev/docs/api#createstitches) and ship our own themed implementations of the entire [Stitches API]
- No need for wrapping our components in a `<ThemeProvider>` because they're themed out of the box
- A much simpler API that's similar to styled-components (minus the tagged template literals) but incorporates themed CSS props a la Theme UI
- First-class [variant support][variants] that's declarative and super flexible
- More performant TypeScript types and improved DX

[stitches]: https://stitches.dev/
[stitches api]: https://stitches.dev/docs/api
[variants]: https://stitches.dev/docs/variants


.github/workflows/reports.yml
```diff
@@ -9,5 +9,5 @@ jobs:
       - uses: preactjs/compressed-size-action@v2
         with:
           repo-token: ${{ github.token }}
-          pattern: 'packages/*/dist/**/*'
+          pattern: 'packages/*/dist/**/*.{js,mjs,css,svg}'
```
shawnbot, line 12: This excludes `.map` and `.d.ts` files (and any other, exotic file formats that we may add later) from our size reports.

packages/react/.eslintrc.js
```diff
@@ -24,7 +24,9 @@ module.exports = {
       files: ['**/*.ts{,x}'],
       // these are not needed in TypeScript
       rules: {
+        'no-redeclare': ['off'],
         'no-undef': ['off'],
+        'no-unused-vars': ['off'],
         'import/named': ['off']
```
shawnbot, line 30: I've been disabling rules one by one as they pop up, but if we're going to use the TypeScript ESLint plugin more broadly we'll need to be more systematic about it. Maybe the top-level config only activates the TS parser + plugin for `.ts` and `.tsx` files? 🤔

packages/react/rollup.config.mjs
```diff
@@ -5,11 +5,13 @@ import peerDepsExternal from 'rollup-plugin-peer-deps-external'
 import resolve from '@rollup/plugin-node-resolve'
 import { terser } from 'rollup-plugin-terser'
 import typescript from '@rollup/plugin-typescript'
+import pkgJson from './package.json' assert { type: 'json' }
```
shawnbot, line 8: I'd never seen [import assertions](https://v8.dev/features/import-assertions) before, but Node and ESLint both complained without it in this case. Node now throws a warning import assertions being experimental, so it might be safer to just read and parse the JSON the old-fashioned way.

packages/react/src/constants.ts
```diff
@@ -1,7 +1 @@
-export const SIMULATED_HOVER_CLASS = '__hover__'
-export const SIMULATED_FOCUS_CLASS = '__focus__'
-export const SIMULATED_HOCUS_CLASS = '__hocus__'
-
-export const HOVER = `&:hover, &.${SIMULATED_HOVER_CLASS}, &.${SIMULATED_HOCUS_CLASS}`
-export const FOCUS = `&:focus, &.${SIMULATED_FOCUS_CLASS}, &.${SIMULATED_HOCUS_CLASS}`
-export const HOCUS = `&:hover, &:focus, &.${SIMULATED_HOVER_CLASS}, &.${SIMULATED_FOCUS_CLASS}, &.${SIMULATED_HOCUS_CLASS}`
```
shawnbot, line 7: I've done away with all the stuff used to implement simulated focus states because either [Yagni](https://martinfowler.com/bliki/Yagni.html) or (more likely) somebody smarter has invented a more elegant solution that we can use when we get to the point where hover/focus styles become important.

packages/react/src/components/Text.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const Text = styled('div', {
+  m: 0,
+  // Chinese text should be rendered in Noto Sans TC,
+  // with bold font weights at 500 (not 600)
+  '&:lang(zh)': {
+    fontFamily: '$chinese',
+    $fontWeights$bold: '$fontWeights$boldChinese',
+    letterSpacing: '0 !important'
+  },
```
shawnbot, line 19: I think this is really cool. We've had challenges styling Chinese text on sf.gov because our typography styles are very particular: we should display all Chinese text in Noto Sans TC, _and_ bold text styles of Chinese text need to have a different font-weight (500 instead of Rubik semibold's 600). This solves it quite elegantly, IMO:

- `&:lang(zh)` matches Text components whose own `lang` attribute _or_ that of the closest ancestor is `zh`
- `fontFamily: '$chinese'` → `font-family: var(--sfgov-fonts-chinese)`
- `$fontWeights$bold: '$fontWeights$boldChinese'` → `--sfgov-fontWeights-bold: var(--sfgov-fontWeights-boldChinese`, which overrides the bold font weight for any text style! 🪄

packages/react/src/theme/colors.ts
```diff
@@ -1,125 +1,34 @@
...
+  redL4: '#9b3921',
+  slateL1: '#eff3f4',
+  slateL2: '#5a7a92',
+  slateL3: '#1d4d70',
+  slateL4: '#002b48',
+  yellowL1: '#f8f1df',
+  yellowL2: '#f9e3a3',
+  yellowL3: '#f4c435',
+  yellowL4: '#e0a81a'
+}
```
shawnbot, line 34: Note: I nixed the sequential color scales from this for now, but we can easily bring them back.

PR #140, opened 2022-12-06
React icons!

shawnbot:
This PR introduces first-class SVG icons pulled from Figma. There are a couple of moving parts here:

### `@sfgov/icons`
This new package contains only the unstyled SVG icon assets exported from Figma in both raw `.svg` and `.jsx` component formats, as well as JSON indices that list them. These are all published to npm, as you can see [on unpkg](https://unpkg.com/browse/@sfgov/icons@0.0.0-6c0bf48/).

The Figma import logic lives entirely in [.figmaexportrc.js](https://github.com/SFDigitalServices/design-system/blob/feature/react-icons/packages/icons/.figmaexportrc.js), which is a configuration file for the [figma-export](https://figma-export.marcomontalbano.com/) tool. The files in the `svg` and `react` directories are rebuilt when you run:

```sh
npm run fetch
# which runs...
env-cmd ../../.env npx figma-export use-config
```

Which uses [env-cmd] to load environment variables from the repo root's `.env` file. And in order for _that_ to work, you need to copy [.env.sample](https://github.com/SFDigitalServices/design-system/blob/9b4699638e82a1137cc2c06ab8268e76564fc70b/.env.sample) and replace `YOUR_FIGMA_TOKEN_HERE` with your own [Figma access token].

### `SVGIcon` component
The SVGIcon component is the vehicle for our icons' styles (really just `display: inline-flex`). It's exported from `@sfgov/react` so that we can style other (custom, not-yet-released) SVG icons the same way, but probably doesn't need to be documented in its own right.

### `Icon<name>` components
Every icon's React component is imported from `@sfgov/icons/react`, wrapped as `Component => props => <SVGIcon {...props} as={Component} />`, and re-exported with the original symbol name. I've built a table of the components in Storybook that also acts as documentation, but we'll probably want to move this to Docusaurus so that it's easier to find. The naming scheme for the icon components is `Icon<name>`, where `<name>` is the [PascalCase] version of the icon's name in Figma (`accessibility` → `Accessibility`; `arrow up` → `ArrowUp`, etc.). E.g.

```tsx
import { IconAccessibility } from '@sfgov/react'
export default IconAccessibility
```

As we [discussed in Slack](https://sfdigitalservices.slack.com/archives/CEERV5DCG/p1670970477017929) (sorry, private), this not only makes for friendlier bundle sizes but also improves developer experience in IDEs that autocomplete `import` statements. In VS Code, typing `Icon` in your `import { } from '@sfgov/react` statement gives you an alphabetical list of the icon components:

If we decide to rename icons or want to create aliases, we can do that by just [re-exporting](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#re-exporting_aggregating) them with different names:

```js
export {
  IconAccessibility as IconA11y,
  IconArrowUp as IconUp,
  IconArrowDown as IconDown
} from './icons'
```

The move from a single `<Icon>` component with a `symbol` string prop (which matches the `<sfgov-icon>` custom element implementation) to the named export approach was done in #142.

#### ⚠️ Hacks
The wrapping and re-exporting of the unstyled SVG icons is currently done in a kind of hacky way, using a [JS script](https://github.com/SFDigitalServices/design-system/blob/9b4699638e82a1137cc2c06ab8268e76564fc70b/packages/react/scripts/wrap-icons.mjs) that reads `@sfgov/icons/index.json` and generates TSX from a tagged template literal. The script lints the resulting code with ESLint and _should_ fail if it doesn't parse or can't be auto-fixed, so I think that's a _good enough_ guard against generating invalid code. 😁 

<details>
<summary><b>An outdated note on bundling and tree-shaking</b></summary>
<div>

Currently this bundles all of the icons in the react package, adding up to 22.8K if you import the `Icon` component or the `iconsIndex` object. This is so that we can map the string symbol names (e.g. `'accessibility'`) to the SVG components with a single import:

```jsx
import { Icon } from '@sfgov/react'
export const A11y = () => <Icon symbol='accessibility' />
export const Toggle = ({ up }) => <Icon symbol={up ? 'chevron-up' : 'chevron-down' } />
```

</div>

[env-cmd]: https://github.com/toddbluhm/env-cmd
[figma access token]: https://help.figma.com/hc/en-us/articles/8085703771159-Manage-personal-access-tokens
[pascalcase]: https://techterms.com/definition/pascalcase


storybook/stories/components/Icons.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+  component: Icon,
+  args: {
+    symbol: 'accessibility',
+    color: 'slateL4',
+    width: 20
+  },
+  argTypes: {
+    symbol: {
+      name: 'Symbol',
+      options: sortedKeys,
```
jacksongonzales, line 29: currently this shows up as the index of each entry, would the `iconsByName` export work here instead?

jacksongonzales (approved):
this looks great!! just one small comment on the storybook portion

PR #143, opened 2022-12-14
Use Lerna to publish from Actions

shawnbot:
I ran into a snag trying to install a canary release of `@sfgov/react` from #140: the package references `@sfgov/icons` at `1.0.0`, but that version hasn't been published yet, and won't until we commit the changes to `main`. I think what we want is a more scalable monorepo publishing workflow, and [@lerna-lite/publish] does that out of the box. To replicate the behavior of the primer/publish action we were using before, I've set up conditional steps in the CI workflow:

1. If `github.ref == 'ref/heads/main'`, just run `lerna publish --yes` with the current versions.
2. If `github.ref` starts with `ref/heads/release-`, run `lerna publish --yes prerelease --dist-tag next`, which should just append a prerelease identifier to the current versions.
3. Otherwise, if neither of those steps ran, we run `lerna publish --yes --dist-tag canary "0.0.0-${GITHUB_SHA::7}`, which publishes _all_ of the packages to canary versions like `0.0.0-9d36701`.

[@lerna-lite/publish]: https://www.npmjs.com/package/@lerna-lite/publish


.github/workflows/ci.yml
```diff
@@ -19,29 +19,14 @@ jobs:
...
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+      - run: npm run build -w @sfgov/react
+      - run: echo "NPM_AUTH_TOKEN=${{ secrets.NPM_AUTH_TOKEN }}" >> $GITHUB_ENV
+      - if: github.ref == 'refs/heads/main'
+        run: lerna publish --yes
+        id: publishMain
+      - if: startsWith(github.ref, 'refs/heads/release-')
+        run: lerna publish --yes prerelease --dist-tag pre
```
shawnbot, line 29:
```suggestion
        run: lerna publish --yes prerelease --dist-tag next
```

PR #144, opened 2022-12-15
React global CSS and server-side rendering

shawnbot:
This PR proposes some components and helper functions for generating:

1. Fast-loading Google Fonts stylesheet with `<link rel="preconnect">` and `<link rel="preload" as="style">`, per [this StackOverflow comment](https://stackoverflow.com/a/62219966/20014650). These only make sense to output server-side, since adding them via client-side JS is too late for the stylesheets to preload. We need this to work for Next.js, Storybook, and Docusaurus 😅 

2. Other global styles that we may need, such as `* { box-sizing: border-box; }`. These can load at any point, but should ideally be applied before other CSS and before any of our components render.

3. Critical CSS for server-rendered React components. This is necessary in both Next.js and Docusaurus to prevent the dreaded Flash of Unstyled Content (FOUC) that comes from server-rendered markup styled during client-side [hydration](https://blog.saeloun.com/2021/12/16/hydration.html).

I had to throw a lot at the wall to see what stuck for Next.js SSR (server-side rendering), but it seems to work and _should_ work equally well in Docusaurus. 🤞 You'll see that for Storybook I had to add a `previewHead()` function that includes the requisite link tags for Google Fonts preloading with a helper function that outputs the proper URL for our fonts. (~Currently these are decoupled from our font styles, but I'd like to integrate them more tightly.~ You can see how I consolidated these in #145)

⚠️ This PR includes a "breaking" change (not really in semver because we haven't released anything!) to the font stack in #145.


jacksongonzales (approved):
this looks great! at some point we'll have to dig into customizing fonts on the MDX pages on docusaurus, but we can handle that in a separate PR

PR #146, opened 2022-12-22
We're pivoting to Stitches

jacksongonzales:
## We switched from Styled Components to Stitches! 

After encountering some challenges with Typescript and finding the DX less than great, we tried out [Stitches](https://stitches.dev/). So far, we're finding it a much better experience. It works nicely with our theming needs, and gives us a `styled('<tag>', { ... })` interface that's quite similar to the Styled Components approach.


packages/react/src/icons.tsx
```diff
@@ -1,6 +1,6 @@
 
 import * as icons from '@sfgov/icons/react'
-import { createStyledIcon as wrap } from './components'
+import { createStyledIcon as wrap } from '../src/components/SVGIcon'
```
shawnbot, line 3: This import path is a bit wonky:
```suggestion
import { createStyledIcon as wrap } from './components/SVGIcon'
```

packages/react/src/components/Tile.tsx
```diff
@@ -0,0 +1,93 @@
...
+    flexDirection: 'row'
+  }
+})
+
+const ContentWrapper = styled('div', {
+  p: 20
+})
+
+const TileTitle = styled(TitleXs, {
+  marginBottom: 20
```
shawnbot, line 43:
```suggestion
  mb: 20
```

packages/react/src/components/index.ts
```diff
@@ -4,6 +4,13 @@ export * from './Container'
 export * from './Flex'
 export * from './GlobalStyle'
 export * from './Grid'
+export * from './pageComponents/AboutSection'
+export * from './pageComponents/AgencyHeader'
+export * from './pageComponents/ArchiveInformation'
+export * from './pageComponents/Section'
+export * from './pageComponents/ContactSection'
+export * from './pageComponents/Spotlight'
```
shawnbot, line 12: Just spitballing here because `componets/pageComponents` reads a bit awkwardly: Would a `src/components/page` or `src/pages` directory be better, and then we could add an index that exports all of them so they're easier to move around?

shawnbot (approved):

PR #147, opened 2023-01-06
Feature/tokens

joni-f-g:
### A proof of concept for using [Style Dictionary](https://amzn.github.io/style-dictionary/#/) to generate tokens for front ends to consume


packages/react/src/theme/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { colors } from './colors'
+import * as colors from 'sfgov-design-system/src/tokens/style-dictionary/colors'
```
jacksongonzales, line 1: we'll want to wait to merge this till we settle on our new package structure; currently, this breaks since there's no `sfgov-design-system` package anymore :)

jacksongonzales: plus we can clean up the imports and not have such a long, winding path to get the tokens

PR #157, opened 2023-01-26
Fix website builds

shawnbot:
This fixes the website build partially broken in #153. 😁


package.json
```diff
@@ -35,6 +35,7 @@
     "build:storybook": "wireit",
     "build:urls": "wireit",
     "build:website": "wireit",
+    "build:website-index": "wireit",
```
jacksongonzales, line 38: it appears there's no entry in the wireit graph for this; is a step missing?

PR #162, opened 2023-02-09
Box block, VariantProps fix

shawnbot:
Two things:

1. I set `display: block` in the default styles for `<Box>`, which simplifies `<Box as='a'>` or any other element that isn't block-level by default.
2. I replaced our broken `VariantProps` type with the one from Stitches. Let's see if this improves autocompletion and errors with the `as` prop in sfgov-next. 🤞

### DIFF
--- a/react/src/components/Flex.tsx
+++ b/react/src/components/Flex.tsx
@@ -13,4 +13,7 @@ export const Flex = styled(Box, {
   }
 })
 
-export type FlexVariant = VariantProps<typeof Flex>
+export type FlexProps = VariantProps<typeof Flex>
+
+// FIXME[1.0]: remove this alias
+export type FlexVariant = FlexProps
--- a/react/src/utils.tsx
+++ b/react/src/utils.tsx
@@ -1,8 +1,8 @@
 import React, { ComponentProps, ComponentType } from 'react'
 
-export type { VariantProps } from '@stitches/react'
+// XXX: export type { VariantProps } from '@stitches/react'
 
-// export type VariantProps<T> = JSX.IntrinsicAttributes & (T extends ComponentType ? ComponentProps<T> : T)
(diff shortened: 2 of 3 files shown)


react/src/components/Flex.tsx
```diff
@@ -13,4 +13,7 @@ export const Flex = styled(Box, {
   }
 })
 
-export type FlexVariant = VariantProps<typeof Flex>
+export type FlexProps = VariantProps<typeof Flex>
+
+// FIXME[1.0]: remove this alias
+export type FlexVariant = FlexProps
```
shawnbot, line 19: `FlexVariant` was a typo 😭 so I'm adding `FlexProps` and aliasing it to `FlexVariant` so that we don't have to release this as a breaking change. The `FIXME[1.0]` comment _should_ remind us to remove the alias when we release 1.0.

react/src/utils.tsx
```diff
@@ -1,4 +1,8 @@
-import React, { Component, ComponentProps, ComponentType } from 'react'
+import React, { ComponentProps, ComponentType } from 'react'
+
+// XXX: export type { VariantProps } from '@stitches/react'
```
jacksongonzales, line 3: should we just delete this? also, I think I'm confused on whether the stitches `VariantProps` is working or causing issues. but down to merge this and see if it fixes the linting errors

joni-f-g (approved):

react/src/utils.tsx
shawnbot: I don't know, tbh. It doesn't hurt to keep the comment around as a reminder that we could bring back the stitches type if/when it's fixed upstream, though.

jacksongonzales: sgtm!

